PR #10, opened 2018-04-03
Integ test mod

jake-smith-hdfgroup:
In-progress but stable revision of "integ" tests `/tests/integ/*`. Adds in-progress swagger documentation `/swagger.yaml`. Test re-write is to make the tests easier to read and modify, as well as expand to cover edge cases of the REST API. Also working towards goal of separating API behavior tests from host-DN-SN operations, and isolating extant tests that rely on an uploaded file.


bilalshaikh42:
@jreadey @jake-smith-hdfgroup I assume this is now outdated. Please open if not

PR #21, opened 2020-02-10
Azure

pkarbhari:
Added Azure related docs and AKS (Kubernetes) .ymls


jreadey:
Docs look good.
I tried to follow this myself, but had a problem at the first step:

$curl -L https://aka.ms/InstallAzureCli | bash

Got this error:
`
Downloading Azure CLI install script from https://azurecliprod.blob.core.windows.net/install.py to /tmp/azure_cli_install_tmp_Q2XYjU.
######################################################################## 100.0%
/tmp/azure_cli_install_tmp_Q2XYjU: OK
Running install script.
-- Verifying Python version.
-- Python version 3.7.4 okay.
-- Verifying native dependencies.
-- Executing: 'rpm -q gcc libffi-devel python37u-devel openssl-devel'
-- One or more of the following native dependencies are not currently installed and may be required.
"yum check-update ; yum install -y gcc libffi-devel python37u-devel openssl-devel"

===> Missing native dependencies. Attempt to continue anyway? (y/N):`

Should we be doing a pip install for az-cli?

pkarbhari:
Thanks. I will make the updates. You will need to run the following 2 commands. I had done my testing on Windows and testing on Linux was done on a machine that already had pip.
sudo apt-get update && sudo apt-get -y upgrade
sudo apt-get install python3-pip

jreadey:
I generally install python packages using a conda environment like miniconda (https://docs.conda.io/en/latest/miniconda.html).  This lets people install packages without sudo privileges or messing up system files.  

With miniconda I can install azure-clie with: `$pip install azure-cli`.

jreadey:
I got an error creating the container.  Looks like you need to use "--" ass a prefix to the storage name.  i.e. not:
`$ az storage container create -n $CONTAINERNAME account-name $STORAGEACCTNAME`
but:
`$ az storage container create -n $CONTAINERNAME --account-name $STORAGEACCTNAME`
That created the container, but got this message:
`No connection string, account key or sas token found, we will query account keys for your storage account. Please try to use --auth-mode login or provide one of the following parameters: connection string, account key or sas token for your storage account.
{
  "created": true
}`

jreadey:
For the kubernetes install...

Rather than the connection string being specified in the deployment yaml, it would be better is kubernetes secrets were used (similar to the access keys for the AWS deployment).

pkarbhari:
Ok, I believe that it should be ready to go. There are changes to Kubernetes_install_azure.md from Kate and I + added Kubernetes secrets for Azure Connection String. Kate is going to run through it one more time after this last commit

jreadey:
In the docker install, test 15 says: "Go to http://hsds.hdf.test/about...", but it should be whatever the HSDS_ENDPOINT is set to.  Also,  should clarify the action is to put the url in a browser.

jreadey:
In docker install, before running testall.py, you'll need to set the admin password: `$export ADMIN_PASSWORD=admin` or whatever it was set to.

jreadey:
In the docker install - "pip install numpy" is needed to run all the integration tests.
Also with numpy, you can just run: "python testall.py" rather than "python testall.py --skip_unit"

PR #29, opened 2020-03-14
Add authentication expiration

jjaraalm:
Re-authentication is required every `AUTH_EXPIRATION` seconds if this is set and > 0. By default, no re-authentication is required in order to preserve old behavior. Credentials set in the `passwd.txt` file never expire.

I did not see any authentication tests, so if they exist they have not been updated. Kubernetes deployment files have also not been updated since I am not as familiar with it.

### DIFF
--- a/runall.sh
+++ b/runall.sh
@@ -37,11 +37,6 @@ echo "Using compose file: ${COMPOSE_FILE}"
 
 [[ -z ${HSDS_ENDPOINT} ]] && echo "HSDS_ENDPOINT is not set" && exit 1
 
-# Do not allow credentials to expire by default.
-if [[ -z ${AUTH_EXPIRATION} ]] ; then
-  export AUTH_EXPIRATION="-1"
-fi
-
 if [[ -z ${PUBLIC_DNS} ]] ; then
   if [[ ${HSDS_ENDPOINT} == "https://"* ]] ; then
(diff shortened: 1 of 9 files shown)


runall.sh
```diff
@@ -37,6 +37,11 @@ echo "Using compose file: ${COMPOSE_FILE}"
 
 [[ -z ${HSDS_ENDPOINT} ]] && echo "HSDS_ENDPOINT is not set" && exit 1
 
+# Do not allow credentials to expire by default.
+if [[ -z ${AUTH_EXPIRATION} ]] ; then
```
jreadey: You shouldn't need to set the default here - it will get picked up from the default in config.py if there's no environment variable set.

jjaraalm: This is actually one thing I did not understand. 

In `config.get()` it looks there's an asymmetry where, for example, `config.get('A')`  will give a return value if `A` in the environment variable, otherwise it will return `None`. On the otherhand `config.get('a')` will return the default value for `a` instead of `None`. It wasn't clear to me if this was intentional or a bug.

Either way, I'll remove this and update.

jreadey:
@jjaraalm - try merging from mainline - that should fix the travis issue.

jjaraalm:
okay, I merged ... but now travis doesn't run? :sweat_smile:

jjaraalm:
Got it. Updating `travis.yml` prevents travis from running in some cases.

PR #31, opened 2020-03-18
Dcos support

s004pmg:
Very early, prototype level support for HSDS on DCOS.


hsds/util/idUtil.py
```diff
@@ -400,7 +402,7 @@ def getDataNodeUrl(app, obj_id):
     node_count = app["node_count"]
     node_state = app["node_state"]
     if node_state!= "READY" or node_count <= 0 or node_count != len(dn_urls):
-        log.info(f"getDataNodeUrl returning 503 - node_state: {node_state} node count: {node_count}")
```
jreadey: Is there any reason not use f-strings here?

hsds/basenode.py
```diff
@@ -450,7 +580,7 @@ def getHeadUrl(app):
 
                     if this_node is None and rsp_json["cluster_state"] != "READY":
                         log.warn("this node not found, re-initialize")
-                        app["node_state"] == "INITIALIZING"
```
jreadey: Thanks for fixing this stupid bug (I'm surprised pyflakes doesn't catch this)

hsds/basenode.py
```diff
@@ -134,10 +148,10 @@ def getHeadUrl(app):
     try:
         await http_post(app, req, data=body)
     except ClientError as client_exception:
-        log.error(f"got ClientError registering with oio_proxy: {client_exception}")
+        log.error(f"got ClientError registering with oio_proxy: {client_exception} and body {body}")
         return
     except CancelledError as cancelled_exception:
-        log.error(f"got CanceeledError registering with oio_proxy: {cancelled_exception}")
```
jreadey: Please change "Canceeled" to "Cancelled"!  (bad spelling in the original code I know)

hsds/basenode.py
```diff
@@ -61,7 +68,14 @@ def getHeadUrl(app):
     req_reg = head_url + "/register"
     log.info("register: {}".format(req_reg))
 
-    body = {"id": app["id"], "port": app["node_port"], "node_type": app["node_type"]}
+    log.warn("env:{}".format(os.environ))
```
jreadey: Why are we dumping os.environ as a warning?

hsds/basenode.py
```diff
@@ -61,7 +68,14 @@ def getHeadUrl(app):
...
+
+    if "is_dcos" in app:
+        outside_port = os.environ.get('PORT0')
```
jreadey: Rather than using os.environ here (note this would throw a KeyNotFound exception is PORT0 is not set) you can use config.get("PORT0").  Alternatively, you may want to put a dict of docs related stuff when you set "is_dcos". e.g. app["dcos_config"] = {'PORT0': 999}, and then refer to that here.

hsds/basenode.py
```diff
@@ -654,6 +784,12 @@ def baseInit(loop, node_type):
         log.info("running in kubernetes")
         app["is_k8s"] = True
 
+    # check to see if we are running in a DCOS cluster
+    is_dcos = os.environ.get('MARATHON_APP_ID')
```
jreadey: Won't this throw an exception for non-dcos?

hsds/util/idUtil.py
s004pmg: No, fixed

hsds/basenode.py
s004pmg: fixed

hsds/basenode.py
s004pmg: Should have been taken out before check-in, this was temporarily useful for seeing what useful things DCOS was given in the environment.

hsds/basenode.py
s004pmg: fixed

hsds/basenode.py
s004pmg: fixed

Issue #34, opened 2020-03-24
Make hsds an application


t20100:
It would be convenient if `hsds` could "scale down" to a simple application so as to use it locally simply as an application without docker.

jreadey:
Are you thinking of something like this: https://github.com/HDFGroup/hsds/blob/master/docs/design/direct_access/direct_access.md?

t20100:
Yes pretty much, except that it could be run as a standalone server rather than being integrated into `h5pyd`, but I guess both can be achieved at once.

jreadey:
This is implemented in PR #35, merged into master.

jreadey:
Hey @t20100 - I'm not that familiar with aiohttp.Web.AppRunner, but it looks like it runs the server within the same process.
I think it would be more efficient to run head, SN, and DN nodes in their own processes so that the server could utilize multiple cores.

What would you think about kicking off a subprocess for each SN, DN node?  (the head node could live in the parent process)

t20100:
Hi,

Yes, as it is everything runs in the same process, I expect it should be simple to spawn subprocesses instead.
But I usually try to avoid subprocesses when I can, what do you think subprocess would improve?

jreadey:
Thanks for confirming...

Here's an example where running in a subprocess would speed things up:

Say you have a SN worker and 4 DN workers.  The SN gets a request to read a dataset selection crossing 10 chunks.  The chunk read requests get spread out over the 4 workers.  Each DN worker needs to fetch the chunk from S3 which is async, so not a big deal to do in one process (I think).  But any uncompression will be hogging the CPU and blocking the other workers.

With docker I observe that we can get close to 100% CPU for each of 4 containers.  For HSDS apps, being able to processes on their own core should be faster (ymmv).

t20100:
OK, I though no lengthy operation was done in either DN or SN loop, but if it is the case, yes, multiprocessing or maybe multithreading would be better.

BTW, when I did it, I set DN and SN count to one when started as application: https://github.com/HDFGroup/hsds/blob/master/hsds/app.py#L67 . That would need to be changed as well.

t20100:
I gave it a try with threads and more or less managed to make it passes the tests for 1 SN and 1 DN, but not with many... Where I stopped, I guessed that each DN should have a different config/env var DN_PORT.

The way the config is handled has changed (now in a yaml file), and the application would need to be updated accordingly.

I can't spend much time on this now, but I can come back on it later.

[...]

PR #35, opened 2020-03-24
Make hsds available as an application

t20100:
This PR builds on PR #26 which content is included here.
It is a first proposition to enable starting `hsds` as a simple application (issue #34).
There is room for improvement, but it passes the tests for both open.io and posix storage.
I takes a few options on top of those `hsds.config.get` is interpreting:

```
% hsds --help
usage: Starts hsds a REST-based service for HDF5 data.

optional arguments:
  -h, --help            show this help message and exit
  --bucket-dir BUCKET_DIR
                        Directory where to store the object store data
  --bucket-name BUCKET_NAME
                        Name of the bucket to use (e.g., "hsds.test").
  --host HOST           Address the service node is bounds with (default: localhost).
  -p PORT, --port PORT  Service node port (default: 5101).
  --s3-gateway S3_GATEWAY
                        S3 service endpoint (e.g., "http://openio:6007")
  --access-key-id ACCESS_KEY_ID
                        s3 access key id (e.g., "demo:demo")
  --secret-access-key SECRET_ACCESS_KEY
                        s3 secret access key (e.g., "DEMO_PASS")
  --password-file PASSWORD_FILE
                        Path to file containing authentication passwords (default: No authentication)

Examples:

- with openio/sds data storage:

  hsds --s3-gateway http://localhost:6007 --access-key-id demo:demo --secret-access-key DEMO_PASS --password-file ./admin/config/passwd.txt --bucket-name hsds.test

- with a POSIX-based storage for 'hsds.test' sub-folder in the './data' folder:

  hsds --bucket-dir ./data/hsds.test
```

### DIFF
--- a/hsds/app.py
+++ b/hsds/app.py
@@ -1,7 +1,6 @@
 import argparse
 import asyncio
 import os
-import sys
 
 from aiohttp import web
 
--- a/hsds/app.py
+++ b/hsds/app.py
@@ -36,8 +36,10 @@ def main():
     args = parser.parse_args()
 
     os.environ['HSDS_ENDPOINT'] = 'http://localhost:5101'
+    os.environ['HEAD_ENDPOINT'] = 'http://localhost:5100'
     os.environ['PUBLIC_DNS'] = 'localhost:5101'
     os.environ['LOG_LEVEL'] = 'DEBUG'
+    os.environ['STANDALONE_APP'] = 'True'
 
     os.environ['TARGET_SN_COUNT'] = '1'
     os.environ['TARGET_DN_COUNT'] = '1'
@@ -46,12 +48,13 @@ def main():
     os.environ['AWS_ACCESS_KEY_ID'] = args.access_key_id[0]
     os.environ['BUCKET_NAME'] = args.bucket_name[0]
     os.environ['PASSWORD_FILE'] = args.password_file[0]
-    
+
     log.info('APP about to start')
-    
+
     from . import datanode, servicenode, headnode
 
     loop = asyncio.get_event_loop()
+
 
     head_runner = web.AppRunner(headnode.create_app(loop))
     dn_runner = web.AppRunner(datanode.create_app(loop))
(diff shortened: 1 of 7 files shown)


t20100:
Rebased as PR #26

jreadey:
Can you rebase to the current master?

t20100:
Rebased to master.

jreadey:
Going back to #26 - I notice that build.sh installs a lot of azure packages each time. E.g. azure-datalake-store.   Is there a way to avoid these?  I thought the required packages where being installed in the Docker base image.  I prefer not having to deal with possible package updates with every build.

t20100:
PR  #49 fixes that by avoiding to check/install dependencies

jreadey:
Do we need anything from this PR or is it all included in PR #49?

t20100:
PR #49 was only to avoid checking the dependencies while installing `hsds`.
This PR turns `hsds` into an application that runs the head, service and data node all at once to ease running `hsds` without docker.

jreadey:
Ok - got it.
It looks like there's a pyflakes error: https://travis-ci.org/github/HDFGroup/hsds.  Could you take a look?

t20100:
Fixed missing coma.

jreadey:
Thx - merged to master.

jreadey:
Hey - is it necessary to copy the entire project tree to the docker image: https://github.com/HDFGroup/hsds/commit/4153b74b2ed8df49adf6e38e5caffdd21fe0e96d#diff-3254677a7917c6c01f55212f86c57fbf?

This may be a security concern for some users.

I made this change were I just delete the src directory: https://github.com/HDFGroup/hsds/commit/5b7a0f1fa1cd67d1fa789099f6bde89a183ca1ba and everything seemed to work fine.

Can pip install work without copying the files over in the first place?

t20100:
Hi,

The source directory is only needed to do the installation, then it is not needed, so deleting it is fine.
I am not really familiar with docker, but if there is a way to "see" the source directory without copying it while building the image it would work the same way.
Another option would be to build a tarball and only copy this. That would be better  because it would not bring in any files that are not supposed to be there.

PR #38, opened 2020-03-31
Fix CORS tests

loichuder:


loichuder:
I get errors in `testGetDomainVerbose` but I think this is due to the POSIX set up. The test does not pass even on `master` with my set up.

jreadey:
I was seeing an error in attr_test.py - but removed the cors check there and everything was allright.

Not sure what your problem could be with testGetDomainVerbose - did you load the tall.h5 test file?

loichuder:
Oops, sorry I missed that one... :(

I indeed have the file loaded and `testGetDomain` succeeds but `testGetDomainVerbose` fails. From my investigations, it is because `getRootInfo` (https://github.com/HDFGroup/hsds/blob/master/hsds/servicenode_lib.py#L244) returns `None` instead of the info JSON. As this method uses `getS3key`, perhaps it does not support POSIX ?

PR #49, opened 2020-04-01
Do not install dependencies when building docker

t20100:
Following https://github.com/HDFGroup/hsds/pull/35#issuecomment-606951083 :
"""
Going back to #26 - I notice that build.sh installs a lot of azure packages each time. E.g. azure-datalake-store. Is there a way to avoid these? I thought the required packages where being installed in the Docker base image. I prefer not having to deal with possible package updates with every build.
"""

This PR removes the check/installation of dependencies when installing `hsds`, so it is expected that all dependencies are already available in the base docker image.


jreadey:
nice!

PR #52, opened 2020-04-03
Google OpenID Support

jjaraalm:
Adds support for authenticating against Google OpenID and potentially other providers. Client support is in https://github.com/HDFGroup/h5pyd/pull/81.

Azure tests should be repeated (I do not have Azure access) and docs need to be updated still.

## New settings
1. `openid_provider`  - provider to use for OpenID authentication. Valid values are `azure` (default) and `google`.
2. `openid_audience` - The intended audience of the OpenID token. If not set, falls back on the `azure_resource_id` variable. For Google, this should be the `client_id`. 
3. `openid_claims` - List of claims to use when resolving usernames from OpenID token in order of priority. The default is the previously hardcoded Azure setting `unique_name,appid`. Neither of these are defined by Google. `sub` would be valid for all OpenID configurations, but is opaque and not helpful for setting ACLs.


jreadey:
Thanks @jjaraalm!  This was working for me with AD.   I made one change: https://github.com/HDFGroup/hsds/commit/6ad25d31b94375bc371105820528d8eb4074f969 since sometimes the config get evaluates to empty string vs None.

PR #61, opened 2020-05-22
Use sdist and install from tarball in docker

t20100:
Following https://github.com/HDFGroup/hsds/pull/35#issuecomment-632186279, this PR proposes to build the source tarball and use it instead of the source directory to build the docker image so as to avoid copying the whole source directory.

`--no-binary hsds` is there to avoid building (and thus storing) the corresponding wheel in the docker image.

This is probably a bit cleaner but I think it has little advantage over the current way: it makes sure only relevant files are used to install `hsds` but at the expense of an extra `python setup.py build` step...

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
   - pip install pyyaml
   - pyflakes hsds/*.py
   - pyflakes hsds/util/*.py
-  - python setup.py build
+  - python setup.py sdist
   - cp admin/config/passwd.default admin/config/passwd.txt
   - docker build -t hdfgroup/hsds .
 script:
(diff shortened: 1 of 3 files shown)


t20100:
Typo fixed, using `sdist` rather that `build`

jreadey:
@t20100 - I think setup.py has evolved so this PR doesn't work as is.  I've made some equivalent changes here: https://github.com/HDFGroup/hsds/commit/0b71becec8a6d6fd1de589ab3601199fc1fbb2c1.  I think this resolves the issue of image size blowing just because I left a 1gb size file in my source tree!

PR #64, opened 2020-07-01
Copy the passwd.txt into the docker container.

imikejackson:
During setup we found that the passwd.txt file was not copied into the docker container. This will fix that issue.

Signed-off-by: Michael Jackson <mjackson@caboose.bluequartz.net>


jreadey:
Haven't heard back from @imikejackson , so closing this PR.

PR #85, opened 2021-04-06
Fix empty value when integers in select param

loichuder:
Putting integers in the `select` query parameter of GET value would give an empty value.

`select=[n]` would indeed be transformed to the empty slice `[n:n]`.

In this PR, `select=[n]` is interpreted as `[n:n+1]`, yielding a slice containing the element `n`.


jreadey:
Looks good, thanks for the PR

PR #95, opened 2021-07-09
Update k8s_make_configmap.sh

Olshansk:
Update path in helper script to reflect the instructions in https://github.com/HDFGroup/hsds/blob/master/docs/kubernetes_install_aws.md.


jreadey (approved):
thx!

Issue #107, opened 2021-11-16
Error regarding FileClient


bilalshaikh42:
Hello,

We are receiving 500 errors for every path that we query, except for `/info`. Our backend is an s3 bucket, that seems to be working fine. 
Here is the error message 

```

6
ERROR> FileClient init: root_dir config not set
5
ERROR> FileClient init: root_dir config not set
4
ERROR> FileClient init: root_dir config not set
3
ERROR> FileClient init: root_dir config not set
2
ERROR> FileClient init: root_dir config not set
1
ERROR> FileClient init: root_dir config not set`
```
We are running tag  v0.7beta8 on K8s, though we had the issue with   v0.7.0beta7 as well. No changes were made to the config  through the cluster was redeployed

Any pointers on how to debug?

jreadey:
Looks like HSDS is trying to use posix storage rather than S3.  Are you setting the AWS_ACCESS_KEY_ID and AWS_SECRET_KEY_ID environment variables?  See the AWS deployment yaml: https://github.com/HDFGroup/hsds/blob/master/admin/kubernetes/k8s_deployment_aws.yml,

jreadey:
Note that the AWS keys are set using the aws-auth-keys secret.  Do you have these set?  `$ kubectl describe  secrets aws-auth-keys`.

Not related to the error you are seeing, but note that in v0.7beta8, the HEAD_PORT env variable needs to be set to null (as in the above yaml).  These is used to toggle between deployments where the pods work together (HEAD_PORT is null) and independently (HEAD_PORT points to a head container on the pod).

bilalshaikh42:
Hello, sorry for the delay. We have the aws-auth-keys secret set. This has not changed recently. 
Here is the config we have currently. 
```\

bilalshaikh42:
# HSDS configuration
allow_noauth: true # enable unauthenticated requests
auth_expiration: -1 # set an expiration for credential caching
default_public: false # new domains are publically readable by default
aws_access_key_id: null # Replace with access key for account or use aws_iam_role
aws_secret_access_key: null # Replace with secret key for account
aws_iam_role: null # For EC2 using IAM roles
aws_region: null
hsds_endpoint: https://data.biosimulations.dev # used for hateos links in response
aws_s3_gateway: http://s3low.scality.uchc.edu # use endpoint for the region HSDS is running in, e.g. 'https://s3.amazonaws.com' for us-east-1
[...]

PR #108, opened 2021-11-17
feat: support config files with .yaml extension

bilalshaikh42:
Motivated by #107. It seems that `.yaml` is the "official" extension, though numerous kubernetes examples use `.yml` leading to confusion. See https://yaml.org/faq.html.


jreadey (approved):
Looks good, thanks for the PR.

Issue #110, opened 2021-11-18
aiobotocore API has changed


mkatgert:
When building the docker image from the master branch I ran into the issue that aiobotocore recently has seen the release of version 2.0, in which the API has changed a bit: get_session() is not in __init__.py anymore but it has moved to aiobotocore.session.  See the [docs](https://aiobotocore.readthedocs.io/en/latest/)

This gives an issue, because the Dockerfile installs the latest version of aiobotocore, and the server code imports get_session from the main package and not from aiobotocore.session. The containers fail to start because of this issue.

Possible solutions would be to convert imports (a lot of them) or to pin to version 1.4.2 of aiobotocore.

jreadey:
This should be fixed with the above merge.  Thanks @benfre!

PR #112, opened 2021-11-20
fix get_session import after aiobotocore 2.0.0

benfre:
In `aiobotocore` 2.0.0 remove `get_session` at top level, as point out in #110 
The fixs are import from `aiobotocore.sessio`


jreadey (approved):
Nice - thanks for the fix.

PR #132, opened 2022-03-29
Automatically tag and push docker images

bilalshaikh42:
This workflow will automatically build the docker image and push it to dockerhub with specific tags as configured in the "Docker meta" step. 

Currently, it is set to use the following tags: 

Push to master -  'master', "GIT_SHA"
Release  (X.Y.Z)- 'X', "X.Y", "X.Y.Z"
 
This can be configured as described here https://github.com/docker/metadata-action

The repository maintainers would need to create two secrets for the repository,   DOCKERHUB_USERNAME and DOCKERHUB_TOKEN to enable pushing the image

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -1,0 +1,52 @@
+name: Docker Image CI
+
+on:
+  push:
+    branches:
+      - 'master'
+    tags:
+      - 'v*'
+  pull_request:
+    branches:
+      - 'master'
+jobs:
+  docker:
+    runs-on: ubuntu-latest
+    steps:
+      -
+        name: Checkout
+        uses: actions/checkout@v2
+      -
+        name: Set up QEMU
+        uses: docker/setup-qemu-action@v1
+      -
+        name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      
+      -
+        name: Docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: hdfgroup/hsds
+          tags: |
+            type=ref,event=branch
+            type=ref,event=pr
+            type=semver,pattern={{version}}
+            type=semver,pattern={{major}}.{{minor}}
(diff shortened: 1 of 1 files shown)


jreadey:
Thanks for the PR!

I don't quite follow how the docker image tag gets set.  These's an HSDS_VERSION string defined here: https://github.com/HDFGroup/hsds/blob/master/hsds/basenode.py#L38.  Is there some way to use that as the tag?

Given a docker tag, is there someone we could to determine the actual commit used for the build?  That would be handy for investigating issues with a specific image.

bilalshaikh42:
There are a few different tags that are set. 
On a push to the master branch, the image will have the tags "master" and the commit hash. The tag "master" would be overridden on the next push, but the next push would have a new commit hash tag. 

The version tags would only be pushed when a release is made on GitHub. So to release the "V0.7.0beta13" tag to dockerhub, you would just create a release on the GitHub repo with that same tag.  The actual commit hash of the release should also be pushed. 

The action is designed to use the git repositories metadata rather than look at the actual code. So rather than reading the file mentioned the approach would be to make a release on Github that would then update that file through some other workflow.

bilalshaikh42:
@jreadey Would this be something you are interested in including? This should reduce the friction of getting new releases to dockerhub. We are also running into an issue where an image is no longer present. Did the beta12 image get removed?

jreadey:
@bilalshaikh42 - yes it is, I've been just to busy to get around to it.  Actually would like to have github actions for linting, build, PyPI push, tests, etc.  Haven't had automatic build tests since Travis went away.  Anyway, I'll try to get to it soon.

I accidently deleted the beta12 tag last week.  Pushed a new beta12 tag just now.

jreadey:
That would be great!  Add sent you an invite as github collaborator.  Email me if you need access to dockerhub, PyPi, etc.
 
For the test integration, if docker is problematic, you can use some variant of "runall.sh --no-docker" to start the service.

PR #135, opened 2022-04-22
Lambda socket dir

PjEdwards:
Fixes an issue with lambda that was introduced about a month ago with changes supporting Windows (I think). Those changes removed the auto-generated socket directory. This reintroduces it in the lambda function only. I'm not at all confident that this is where you'd want this, and will be glad to modify if you have some feedback about a proper place to put it per your coding standards.


jreadey (approved):
Thanks Paul - I had done a bunch of refactoring with HsdsApp but hadn't checked to see how that worked with Lambda.

PR #140, opened 2022-04-29
Fix missing argument in log.info() in getDomainObjects()

holzman:


jreadey (approved):

jreadey:
Thanks for catching this.

PR #153, opened 2022-05-30
build: create new docker image

bilalshaikh42:


jreadey:
Strange - my image size is much larger.  1.12 GB vs 493 MB.
Also container fails at startup: /bin/bash: line 1: entrypoint.sh: command not found

Not sure what's going on.

jreadey:
Could you add a comment in the Dockerfile showing how to generate a new pipfile?

PR #157, opened 2022-06-30
Allow use of unsigned requests to S3

mbruggs:
I'm working on a project in which the S3 buckets are public and therefore don't require any AWS credentials. The commit adds a new config option to achieve the same behaviour that AWS CLI gives with `--no-sign-request`.

Please let me know if there is anything that you think could be done better.

Thanks again for the work on HSDS.


jreadey:
Nice!  One suggestion: can you update to use the "default" option in config.get()?   This will avoid problems if anyone is running with an older version of the config.yml.

mbruggs:
Good catch, thanks. Updated with False as the second argument.

PR #191, opened 2022-10-29
Fixes

trmt:
* domain_sn.py - fix blosc.get_nthreads() call
* util/storUtil.py - remove extra log.warn() with undefined variable "msg"


jreadey (approved):

trmt:
It was just test run in systemd-nspawn container with cpu limit option. that's why i set blosc_nthreads to zero and caught that error

jreadey:
Good you found that.  Setting blosc_nthreads to zero may not be what you want to limit cpu usage though.  Zero causes the number of threads to be based on whatever the processor supports.  Using a value of one would be the most cpu-limiting choice.

Issue #204, opened 2023-01-09
Connection refused when running with docker on Azure VM


wiwski:
Hi,

I am having some issues trying to setup hsds on an Azure VM following this guide : https://github.com/HDFGroup/hsds/blob/master/docs/docker_install_azure.md

I just can't access $HSDS_ENDPOINT/about as I am getting the following error :
```
curl: (7) Failed to connect to euphro-stg-hsds.westeurope.cloudapp.azure.com port 80 after 8 ms: Connection refused
```

Port 80 is opened in networking rules, DNS name is set as well.

Here is the ouput of `docker ps` :
```
CONTAINER ID   IMAGE           COMMAND                  CREATED          STATUS          PORTS                                                                NAMES
0f313bf0e070   hdfgroup/hsds   "/bin/bash -c /entry…"   11 seconds ago   Up 9 seconds    5100-5999/tcp, 0.0.0.0:49158->6101/tcp, :::49158->6101/tcp           hsds_dn_1
0b6f50bd55ec   hdfgroup/hsds   "/bin/bash -c /entry…"   11 seconds ago   Up 9 seconds    5100-5999/tcp, 0.0.0.0:49157->6101/tcp, :::49157->6101/tcp           hsds_dn_3
bf05f21c11ae   hdfgroup/hsds   "/bin/bash -c /entry…"   11 seconds ago   Up 9 seconds    5100-5999/tcp, 0.0.0.0:49156->6101/tcp, :::49156->6101/tcp           hsds_dn_2
012f3162bdfb   hdfgroup/hsds   "/bin/bash -c /entry…"   11 seconds ago   Up 9 seconds    5100-5999/tcp, 0.0.0.0:49155->6101/tcp, :::49155->6101/tcp           hsds_dn_4
c8ef38713535   hdfgroup/hsds   "/bin/bash -c /entry…"   12 seconds ago   Up 11 seconds   5100-5999/tcp, 0.0.0.0:49154->6900/tcp, :::49154->6900/tcp           hsds_rangeget_1
f5ed03ee426b   hdfgroup/hsds   "/bin/bash -c /entry…"   12 seconds ago   Up 11 seconds   5100/tcp, 5102-5999/tcp, 0.0.0.0:5101->5101/tcp, :::5101->5101/tcp   hsds_sn_1
[...]

PR #205, opened 2023-01-09
add instructions on using SN_PORT env variable

jreadey:
Clarify docker setup based on feedback from issue:  #204


jreadey:
Clarify instructions for installing on Docker

PR #219, opened 2023-04-27
Implement path traversal in getObjectIdByPath

mattjala:
Part of a set of changes intended to reduce the number of client requests that need to be made from the REST VOL, specifically HDFGroup/vol-rest#32


jreadey:
@mattjala - could you fix the flake8 errors listed in the python package build?

PR #220, opened 2023-04-27
Groupcreateprop

jreadey:
Add support for datasets created with compact and contiguous layouts.


mattjala (dismissed):

mattjala:
The changes with `H5D_CONTIGUOUS` and `H5D_COMPACT` work fine. 

The `creationProperties` field also works, except in the case of the root group. Because the root group of a domain is created when the domain is created, the body of the request has to be checked for `creationProperties` at domain creation time. So as written, there's no way to set the creation properties for the root group. 

I was able to have the root group inherit a creation property list sent at file creation by adding this after [line 1121 in  domain_sn.py](https://github.com/HDFGroup/hsds/blob/master/hsds/domain_sn.py#L1121)

```
if body and "group" in body and "creationProperties" in body["group"]:
    group_json["creationProperties"] = (body["group"])["creationProperties"]
```

jreadey:
Good catch about the root group property list.  I've checked in your suggested change.

PR #223, opened 2023-05-03
Fix GET datatype including bucket in domain

mattjala:
Responses for other object types (groups, datasets) don't include the bucket, so this makes it more consistent.


jreadey (approved):
looks good.

PR #225, opened 2023-05-08
Linkorder

jreadey:
@mattjala - take a look at this - list by creation order for links or attribute listing.  Are there any vol-tests that require this?


mattjala:
Yes, there are several VOL tests that try out visiting objects by creation order. I'll add support for using this to the REST VOL - it looks good besides my comment about sorting the same list twice.

mattjala (approved):

mattjala (approved):

PR #230, opened 2023-05-31
Chunkinfo

jreadey:
Added a new creationpropertylayout for a "chunk initializer" action.  Curently the only supported action is "hsds-chunklocator".  When set, HSDS will kick of a subprocess as each chunk in the dataset is first accessed.  This process uses h5py to get chunklocations in the referenced HDF5 file and dataset.

See ValueTest.testChunkInitializer to see how this works.

### DIFF
--- a/docs/design/single_object/SingleObject.md
+++ b/docs/design/single_object/SingleObject.md
@@ -115,11 +115,30 @@ Example:
 ```
 "layout": {
         "class": "H5D_CHUNKED_REF_INDIRECT", 
-        "dims": [40, 80],
+        "dims": [100, 100],
         "file_uri": "s3://mybucket/mylocation/myfile.h5",
         "chunk_table": "d-7fbe2e27-87c5e1d8-f736-a6af0f-4d6950"
     }
 ```
+
+By default, the chunk_table needs be initialized by an external program (e.g. `hsload --init`) in 
+order for data in the HDF5 file to be returned correctly.  It is also possible for the chunk table
+dataset to be initialized by HSDS by using the following layout for the chunk_table dataset:
+
+```
+"layout": {
+    "class:": "H5D_CHUNKED",
+    "dims": [10, 10],
+    "chunk_init": "hsds-chunklocator",
+    "init_args": ["s3://mybucket/mylocation/myfile.h5",]
+}
+```
+
+In this example, the chunk table dataset has shape [10,10] and is chunked with one 10 x 10 chunk.  
+When the chunk table dataset if first accessed, the application specified in chunk_init will be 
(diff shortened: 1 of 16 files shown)


docs/design/single_object/SingleObject.md
```diff
@@ -115,12 +115,31 @@ Example:
...
+"layout": {
+    "class:": "H5D_CHUNKED",
+    "dims": [10, 10],
+    "chunk_initializer": "hsds-chunklocator",
+    "initializer_args": ["s3://mybucket/mylocation/myfile.h5",]
+}
+```
+
+In this example, the chunk table dataset has shape [10,10] and is chunked with one 10 x 10 chunk.  
+When the chunk table dataset if first accessed, the application specified in chunk_init will be
```
mattjala, line 138: Nitpick: `if first accessed` -> `is first accessed`

hsds/chunklocator.py
```diff
@@ -0,0 +1,231 @@
...
+    # end chunk_callback
+
+    if dset.shape is None:
+        # Empty (null) dataset...
+        log.warn("Null space dataset")
+        return None
+
+    rank = len(dset.shape)
+    if rank < 1:
+        # no chuunking for scalar datasets
```
mattjala, line 119: Nitpick: `chuunking` -> `chunking`

hsds/util/arrayUtil.py
```diff
@@ -541,3 +541,70 @@ def squeezeArray(data):
...
+        # bump up the last index and carry forward if we run outside the selection
+        dim = self._rank - 1
+        ret_index = self._index.copy()
+        while True:
+            s = self._sel[dim]
+            if s.step:
+                step = s.step
+            else:
+                step = 1
+            s = self._sel[dim]
```
mattjala, line 596: Duplicate line

hsds/util/fileClient.py
```diff
@@ -108,7 +108,20 @@ def _file_stats_increment(self, counter, inc=1):
             log.error(f"unexpected inc for file_stats: {inc}")
             return
 
-        file_stats[counter] += inc
+    def getURIFromKey(self, key, bucket=None):
```
mattjala, line 111: Might be less confusing to have the version here in `fileClient.py` and the version in `azureBlobClient.py` be merged with an argument to decide the exact uri returned.

mattjala (approved):
LGTM

hsds/util/fileClient.py
jreadey: I'll come back to this later - will need some changes for the azure uri strings.

Issue #192, opened 2022-11-14
POST dataset with 0 chunk extent gives 500 error


jreadey:
Values should be verified and a 400 returned for non-positive extents

mattjala:
Resolved in #236

PR #236, opened 2023-06-15
Add test for POST dset with zero chunk extent

mattjala:
Testing the issue reported in #192. 

HSDS already catches invalid chunk extents and returns 400 [here](https://github.com/HDFGroup/hsds/blob/master/hsds/dset_sn.py#L81), so if this elicits the problem correctly, we can close that issue without any further work.


jreadey (commented):
Looks good.
Curious how the 500 in the issue was coming up.

jreadey (approved):

PR #237, opened 2023-06-15
Fix jsonToArray error on empty data read

mattjala:
This lets clients read from an empty attribute without causing an internal server error.

Previously, a value in `data_json` being `None` caused errors when passed to `np.array`.


jreadey (commented):
Could you add a test for this in array_util_test.py?

jreadey (approved):

jreadey (approved):

PR #241, opened 2023-07-31
Fillvalue

jreadey:
Enable base64 encoding for fillvalue.  To use set "fillValue_encoding" in creation properties to "base64"


mattjala (dismissed):

mattjala:
Could you also increment the package version, so that the REST VOL can check HSDS's version to see if fill value encoding is supported?

jreadey:
Change version to 0.8.1

mattjala (approved):

PR #248, opened 2023-08-21
Fix encoded array fill values

mattjala:
Also separates tests for getNumpyValue into encoded/not encoded, and adds array tests for each.


jreadey (approved):
Nice!

PR #266, opened 2023-10-06
Allow PUT shape to reduce extent

jreadey:
Enable PUT shape to reduce extent
Also - fix errors showing up in async process.

### DIFF
--- a/hsds/util/dsetUtil.py
+++ b/hsds/util/dsetUtil.py
@@ -855,48 +855,46 @@ def isExtensible(dims, maxdims):
     return False
 
 
-def getDatasetCreationPropertyLayout(dset_json):
-    """ return layout json from creation property list """
-    cpl = None
+def getDatasetLayout(dset_json):
+    """ Return layout json from creation property list or layout json """
+    layout = None
+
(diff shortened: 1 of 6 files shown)


hsds/util/dsetUtil.py
```diff
@@ -855,48 +855,46 @@ def isExtensible(dims, maxdims):
...
-    return None  # not found
+
+    layout = getDatasetLayout(dset_json)
+    if layout and "dims" in layout:
+        return layout["dims"]
+    else:
+        # H5D_COMPACT and H5D_CONTIGUOUS will not have a dims key
+        # Check the layout dict in dset_json to see if it's
+        # defined there
+        if "layout" in dset_json:
```
mattjala, line 893: Isn't this section redundant, since `getDatasetLayout` already checks for a `layout` directly in `dset_json`?

mattjala (approved):

mattjala:
After decreasing and then increasing size, the new positions in the dataset should have a value equal to fill value, not whatever was written there before the size was decreased. See the modified test in #267

PR #267, opened 2023-10-06
Test decreasing and then increasing size of dset

mattjala:
After decreasing and then increasing size, the new positions in the dataset should have a value equal to fill value, not whatever was written there before the size was decreased.


jreadey:
That behavior will require quite a bit more additional logic.
Let's push this PR to master and then work on the value changes in separate PR.
It will be a good place to include the TBD code to remove chunks that are no longer part of the dataspace.

mattjala:
More comprehensive tests for this are added in #277

Issue #255, opened 2023-08-28
'Verbose' parameters for dset reads returns inaccurate until scan


mattjala:
The 'verbose' parameter for a dataset GET returns information based on all of the datasets' chunks, which is only aggregated whenever a bucket scan occurs. These scans occur [once a minute by default](https://github.com/mattjala/hsds/blob/master/hsds/datanode.py#L101), so requests will return inaccurate info until the next scan occurs. 

This is relevant to HDFGroup/vol-rest#55, since that API call should return the total number of allocated bytes, which must be obtained by the verbose parameter.

jreadey:
Closing this as it should be resolved in master.
Since the task that collect storage statistics runs asynchronously and only looks at chunks in the storage system, clients need to do a domain PUT with rescan and flush params, followed by a dataset GET with a verbose param.  See checkVerbose in tests/integ/value_test.py for an example.

PR #269, opened 2023-10-10
Quickscan

jreadey:
This change addresses issue #255 - verbose info not correct.

For the verbose param to return the current info for domain or datasets, you'll need to do a PUT / with rescan and flush params set.  See the checkVerbose function in value_test.py to see how this works.

### DIFF
--- a/admin/config/config.yml
+++ b/admin/config/config.yml
@@ -32,7 +32,7 @@ log_prefix: null # Prefix text to append to log entries
 max_tcp_connections: 100 # max number of inflight tcp connections
 head_sleep_time: 10 # max sleep time between health checks for head node
 node_sleep_time: 10 # max sleep time between health checks for SN/DN nodes
-async_sleep_time: 10 # max sleep time between async task runs
+async_sleep_time: 1 # max sleep time between async task runs
 s3_sync_interval: 1 # time to wait between s3_sync checks (in sec)
 s3_age_time: 1 # time to wait since last update to write an object to S3
 s3_sync_task_timeout: 10 # time to cancel write task if no response
--- a/hsds/datanode.py
+++ b/hsds/datanode.py
@@ -98,8 +98,8 @@ async def bucketScan(app):
     log.info("bucketScan start")
 
     async_sleep_time = int(config.get("async_sleep_time"))
-    scan_wait_time = async_sleep_time * 6  # default to ~1min
-    log.info("scan_wait_time: {}".format(scan_wait_time))
+    scan_wait_time = async_sleep_time  # default to ~1min
+    log.info(f"scan_wait_time: {scan_wait_time}")
 
     # update/initialize root object before starting node updates
 
--- a/hsds/datanode.py
+++ b/hsds/datanode.py
@@ -98,8 +98,10 @@ async def bucketScan(app):
     log.info("bucketScan start")
 
     async_sleep_time = int(config.get("async_sleep_time"))
+    short_sleep_time = float(async_sleep_time) / 10.0
     scan_wait_time = async_sleep_time  # default to ~1min
     log.info(f"scan_wait_time: {scan_wait_time}")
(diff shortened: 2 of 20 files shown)


hsds/datanode.py
```diff
@@ -165,8 +167,16 @@ async def bucketScan(app):
                 tb = traceback.format_exc()
                 print("traceback:", tb)
 
-        log.info(f"bucketScan - sleep: {async_sleep_time}")
-        await asyncio.sleep(async_sleep_time)
+            last_action = time.time()
+
+        now = time.time()
+        if (now - last_action) > async_sleep_time:
```
mattjala, line 173: If `last_action` is set to the current time before the scanning loop is performed/at the end of each scan loop, won't this always be true?

hsds/datanode.py
```diff
@@ -165,8 +167,16 @@ async def bucketScan(app):
...
+            sleep_time = async_sleep_time  # long nap
+        else:
+            sleep_time = short_sleep_time  # shot nap
```
mattjala, line 176: What's the idea behind varying the time between scans based on how long the scan itself takes? Or is `last_action` tracking something else?

admin/config/config.yml
```diff
@@ -32,7 +32,7 @@ log_prefix: null # Prefix text to append to log entries
 max_tcp_connections: 100 # max number of inflight tcp connections
 head_sleep_time: 10 # max sleep time between health checks for head node
 node_sleep_time: 10 # max sleep time between health checks for SN/DN nodes
-async_sleep_time: 10 # max sleep time between async task runs
```
mattjala, line 35: Why the change in automatic scan frequency? I thought the idea was to allow certain requests to force a scan.

mattjala (approved):
This enabled `H5Dget_storage_size` to work from the REST VOL, when it requests a flush and rescan first. I don't fully understand the reasoning behind all the scan changes, but we can iron those out later. Approved.

hsds/datanode.py
jreadey: If there are no root_ids, last_action doesn't get updated, so eventually the if case will get executed.

hsds/datanode.py
jreadey: If there's nothing going on, use a longer sleep and save cpu cycles.  It's not how long the scan takes but rather the last time there was a domain to scan.

jreadey: Here's a quick check using the value_test.py test case.
I modified the testPut1DDataset to do a sequence of flushes and rescans:

        # check values we should get from a verbose query
        import time
        ts = time.time()
        for i in range(10):
            expected = {"num_chunks": 1, "allocated_size": 40}
            self.checkVerbose(dset_id, headers=headers, expected=expected)
            now = time.time()
            elapsed = now - ts
            print(f"flush {elapsed:.2f}")
            ts = now

Output is:

          $ python value_test.py ValueTest.testPut1DDataset
          testPut1DDataset /home/test_user1/hsds_test/valuetest/20231010T154859_160988Z
          flush 2.54
          flush 0.12
          flush 0.12
          flush 0.13
          flush 0.13
          flush 0.13
          flush 0.12
          flush 0.12
          flush 0.13
          flush 0.13


The first round takes a couple of seconds because apparently  the bucketScan task is in long sleep mode.  After the first flush, it "wakes up" and each update only takes about 0.1s.

admin/config/config.yml
jreadey: The buckeScan task is still running independently, so the sleep is there to not unnecessarily consume cpu.
It might be better to use something like asyncio.Event (see https://docs.python.org/3/library/asyncio-sync.html#asyncio.Event).

jreadey: I'll leave using signals  as a future enhancement since what we have should be good enough for this non-critical path.

Issue #270, opened 2023-10-10
Support fixed-width unicode strings 


mattjala:
HSDS currently does not support these (see `hdf5dtype.py:617`)

ajelenak:
Yes, string encoding and how many bytes are reserved for its storage are decoupled.

mattjala:
> Are they supported in the library?

Yep, see [here](https://github.com/mattjala/vol-tests/commit/a57d0bd6eb91f30a920699b880e817e1ec1a7997) for an example of fixed-length unicode strings being used in datasets/attributes - the native VOL passes both of these tests.

ajelenak:
The question may be more related to how h5py treats HDF5 strings where this combo is not really supported. Any fixed-length string is treated as `bytes` object, not Unicode string.

jreadey:
A fixed width unicode would be utf-32, but like @ajelenak says, it's not explicitly supported by the library. (or HSDS).

mattjala:
> A fixed width unicode would be utf-32, but like @ajelenak says, it's not explicitly supported by the library. (or HSDS).

I think there's a confusion in terminology here. What I'm requesting is not support for a unicode character encoding where each particular character has a fixed width in bytes (e.g. UTF-32), but support for string datatypes that have a fixed total length in bytes (fixed length strings) AND have the character set/encoding UTF-8 (where a particular character does not have a fixed number of bytes associated with it). 

I've updated the title of this issue to be more clear. The library does support fixed-length strings in UTF-8 (See the tests I linked above).

mattjala:
Implemented in #278

PR #278, opened 2023-10-27
add support for fixed width UTF8 strings - #270

jreadey:
Update for UTF8 fixed width strings

### DIFF
--- a/hsds/attr_sn.py
+++ b/hsds/attr_sn.py
@@ -649,6 +649,7 @@ async def PUT_AttributeValue(request):
         raise HTTPBadRequest(reason=msg)
 
     np_shape = getShapeDims(attr_shape)
+    log.debug(f"np_shape: {np_shape}")
     type_json = dn_json["type"]
     np_dtype = createDataType(type_json)  # np datatype
 
@@ -697,6 +698,10 @@ async def PUT_AttributeValue(request):
         # convert to JSON for transmission to DN
         data = arr.tolist()
         value = bytesArrayToList(data)
+        if attr_shape["class"] == "H5S_SCALAR":
+            # just send the value, not a list
+            value = value[0]
+
     else:
         try:
             body = await request.json()
--- a/hsds/chunk_sn.py
+++ b/hsds/chunk_sn.py
@@ -509,7 +509,6 @@ async def PUT_Value(request):
             else:
                 arr = jsonToArray(np_shape, dset_dtype, json_data)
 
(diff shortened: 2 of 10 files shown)


mattjala:
This doesn't seem to work with UTF-8 strings that are sent as base64-encoded binary from the REST VOL. I'll try to get a test set up for this case in python.

mattjala:
It seems that HSDS treats the length-in-characters of the string as its datatype size. Then when a binary request comes in, the length-in-bytes of the same string is seen as being too large for the datatype (if one of the UTF-8 characters is multi-byte). HDF5 considers the `size` field to be length-in-bytes, so HSDS should probably do the same.

mattjala:
Handling requests to write fixed-length UTF8 strings in binary instead of JSON is problematic with how numpy stores unicode strings.

When a client makes a binary write request, HSDS attempts to read the binary buffer into a numpy array with `np.fromstring()` using a numpy datatype that is constructed with `createDataType()`. In the case of a fixed-length UTF8 string datatype, the constructed numpy datatype is `<UXX`, where `XX` is the length of the string datatype. Numpy uses the UTF-32 encoding where each character is always four bytes, so it  expects the string given to `np.fromstring()` to be (about) four times larger than its UTF8 encoding in bytes is, preventing the call from succeeding.

Encoding the given UTF-8 binary to UTF-32 doesn't preserve the size, so the fixed length utf8 strings will no longer have a uniform length in bytes. This prevents `np.fromstring()` from being used to parse the strings into elements of a single fixed-length datatype.

Creating a numpy unicode string datatype with a size that is one fourth the byte-length of the client's UTF-8 bytestring (so that numpy's internal datataype size matches the bytestring's actual size) allows the `np.fromstring()` call to complete, but results in a numpy array with malformed UTF-32 strings that throws an error whenever you attempt to access an element from it. 

This doesn't come up when writing the strings as JSON, since moving the data into the correct shape is handled by `jsonToArray` in that case. 

I'll create a PR with tests to illustrate this issue, though I'm not sure how to resolve it at the moment.

jreadey:
I've added @mattjala binary request tests and fixed some issues with UTF8 encoding...

mattjala:
Running these tests with a fresh environment caused them to pass. It seems that one of my dependencies was outdated, and that was changing the specifics of the encoding. Once the attribute binary test is in, this should be good to merge.

jreadey:
@mattjala - take a look at the revised PR!

PR #281, opened 2023-10-31
Fletcher32

jreadey:
This enables the fletcher32 filter to be set in the creation property list.
Doesn't actually create a checksum but this will be helpful for compatibility with existing apps.

### DIFF
--- a/hsds/util/hdf5dtype.py
+++ b/hsds/util/hdf5dtype.py
@@ -339,12 +339,25 @@ def getTypeItem(dt, metadata=None):
         else:
             # Fixed length string type
             type_info["class"] = "H5T_STRING"
-        type_info["charSet"] = "H5T_CSET_ASCII"
         type_info["length"] = dt.itemsize
+        type_info["charSet"] = "H5T_CSET_ASCII"
         type_info["strPad"] = "H5T_STR_NULLPAD"
     elif dt.base.kind == "U":
         # Fixed length unicode type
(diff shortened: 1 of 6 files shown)


hsds/util/hdf5dtype.py
```diff
@@ -339,12 +339,25 @@ def getTypeItem(dt, metadata=None):
...
+        if ref_check is not None:
+            raise TypeError("unexpected reference type")
+
+        # Fixed length UTF8 string type
+        type_info["class"] = "H5T_STRING"
+
+        # this can be problematic if the encoding of the string is not valid,
+        # or reqires too many bytes.  Use unicode sting length * 4 to handle all
+        # UTF8 strings correctly
+        type_info["charSet"] = "H5T_CSET_UTF8"
```
mattjala, line 357: The fletcher32 support should probably be separated from the fixed width UTF8 string support, since I'm since working on some tests for that

mattjala:
LGTM. In order for the REST VOL's filter tests to fully pass, HSDS would also need to support (or "support" without implementing) the [N-bit filter](https://docs.hdfgroup.org/hdf5/develop/group___d_c_p_l.html#gad58a9c0e766ef71d4075b2c2a755e91c) and the [scale-offset filter](https://docs.hdfgroup.org/hdf5/develop/group___d_c_p_l.html#ga5c10165b670e0e984db431aee818cc7e).

mattjala (approved):

PR #282, opened 2023-10-31
add support for fletcher32

jreadey:
NOP support for fletcher32


mattjala (approved):
LGTM. In order for the REST VOL's filter tests to fully pass, HSDS would also need to support (or "support" without implementing) the [N-bit filter](https://docs.hdfgroup.org/hdf5/develop/group___d_c_p_l.html#gad58a9c0e766ef71d4075b2c2a755e91c) and the [scale-offset filter](https://docs.hdfgroup.org/hdf5/develop/group___d_c_p_l.html#ga5c10165b670e0e984db431aee818cc7e).

PR #284, opened 2023-11-01
Fix jsonToArray on None array

mattjala:
This is the same issue that was partially fixed in #237, where `None` elements in a `value` array caused issues in the helper `jsonToArray`. This covers the case where the entire `value` array is `None` (e.g. an attribute that was created but not written to). It now returns an empty numpy array instead of throwing an exception. 

This also makes HSDS log the error messages `jsonToArray` produces, and adds a try/except block when doing a bytes decode that can fail.

### DIFF
--- a/hsds/attr_sn.py
+++ b/hsds/attr_sn.py
@@ -374,10 +374,13 @@ async def PUT_Attribute(request):
         log.debug(f"attribute value: {value}")
         try:
             arr = jsonToArray(np_dims, arr_dtype, value)
-        except ValueError:
-            msg = "Bad Request: input data doesn't match selection"
-            log.warn(msg)
-            raise HTTPBadRequest(reason=msg)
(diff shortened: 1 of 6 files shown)


jreadey:
Good - not sure what I was thinking with the "if None in data_json...".
Let's keep the HTTP exceptions out of arrayUtil.py.  Would prefer the calling function catch ValueError and then raise HTTP... exception as appropriate.
Also, could you merge in the latest changes from master?

jreadey:
Why do you have the extra parens?  e.g.` if (value is None): ` vs `if value is None:`

mattjala:
> Why do you have the extra parens? e.g.`if (value is None):` vs `if value is None:`

Force of habit from working in C. I'll clean those up.

jreadey (approved):
looks good!

PR #290, opened 2023-11-28
Support creation of array/nested array types

mattjala:
Reading and writing to array/nested array type objects is still buggy, but this is a step in the right direction.

Previously, the check on the return value of getFillValue would throw an error when the datatype was H5T_ARRAY.


jreadey (approved):
It's progress!

PR #292, opened 2023-12-08
Fieldop

jreadey:
Add support for sub-field selection in hyperslab, point, and query operations for both reading and write.
To specify which fields of a compound type are to be accessed, add a query param or body key of "fields".  For a single field selection the query value can just be the field name.  If multiple fields are to be accessed, concatenate the field names with a colon separator.

### DIFF
--- a/hsds/basenode.py
+++ b/hsds/basenode.py
@@ -33,7 +33,7 @@ from .util.authUtil import isAdminUser
 from .util.k8sClient import getDnLabelSelector, getPodIps
 from . import hsds_logger as log
 
-HSDS_VERSION = "0.8.4"
+HSDS_VERSION = "0.8.5"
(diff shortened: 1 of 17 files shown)


mattjala (approved):

mattjala:
CodeQL's warning in `hsds_logger.py` about logging a password seems to be a false positive from printing the username in `authUtil.py:321`. I think should probably remove it anyway to keep the warnings clean.

The warnings in fileClient also want `filepath` to be processed to avoid malicious or unusual paths - that should be a quickfix like in the example [here](https://codeql.github.com/codeql-query-help/python/py-path-injection/).

Besides the CodeQL issues, LGTM.

mattjala (approved):

PR #294, opened 2023-12-18
Support requests with long URL

mattjala:
This allows for users to have link/group names up to about 64Kb, which the main library supports. 

The specific choice of 67584 for the default max size is `65 * 1024 + 2048`. This is because the library's API tests include a link with a name that is `65 * 1024` bytes long, and  2048 bytes should be enough to store the rest of the URL.


mattjala:
After a discussion, it seems like a better solution might be to handle this with something similar to the changes in #295, where long URLs are sent as part of the body of a POST request instead of in the URL.

PR #295, opened 2023-12-18
Multiops

jreadey:
Add ability to get multiple attributes via PUT attributes or write multiple attributes with POST attributes.

### DIFF
--- a/hsds/attr_dn.py
+++ b/hsds/attr_dn.py
@@ -239,23 +239,21 @@ async def GET_Attribute(request):
     return resp
 
 
-async def PUT_Attribute(request):
-    """ Handler for PUT /(obj)/<id>/attributes/<name>
+async def PUT_Attributes(request):
+    """ Handler for PUT /(obj)/<id>/attributes
     """
     log.request(request)
     app = request.app
     params = request.rel_url.query
+    log.debug(f"got PUT_Attributes params: {params}")
     obj_id = get_obj_id(request)
-
-    attr_name = request.match_info.get('name')
-    log.info("PUT attribute {} in {}".format(attr_name, obj_id))
-    validateAttributeName(attr_name)
 
     if not request.has_body:
         log.error("PUT_Attribute with no body")
         raise HTTPBadRequest(message="body expected")
 
     body = await request.json()
+    log.debug(f"PUT Attribute body: {body}")  # remove this to avoid verbosity
     if "bucket" in params:
         bucket = params["bucket"]
     elif "bucket" in body:
@@ -267,68 +265,89 @@ async def PUT_Attribute(request):
     if "replace" in params and params["replace"]:
         replace = True
         log.info("replace attribute")
-    datatype = None
-    shape = None
-    value = None
 
-    if "type" not in body:
-        log.error("PUT attribute with no type in body")
-        raise HTTPInternalServerError()
+    if "attributes" in body:
+        items = body["attributes"]
+    else:
+        # make it look like a dictionary anyway to make
+        # the processing more consistent
+        items = {}
+        if "name" not in body:
+            log.error("PUT attribute with no name in body")
+            raise HTTPInternalServerError()
+        attr_name = body["name"]
+        attribute = {}
+        if "type" in body:
+            attribute["type"] = body["type"]
+        if "shape" in body:
+            attribute["shape"] = body["shape"]
+        if "value" in body:
+            attribute["value"] = body["value"]
+        items[attr_name] = attribute
+    
+    # validate input
+    for attr_name in items:
+        validateAttributeName(attr_name)
+        attr_json = items[attr_name]
+        if "type" not in attr_json:
+            log.error("PUT attribute with no type in body")
+            raise HTTPInternalServerError()
+        if "shape" not in attr_json:
+            log.error("PUT attribute with no shape in body")
+            raise HTTPInternalServerError()
 
-    datatype = body["type"]
-
-    if "shape" not in body:
-        log.error("PUT attribute with no shape in body")
-        raise HTTPInternalServerError()
-    shape = body["shape"]
-
-    if "value" in body:
-        value = body["value"]
+    log.info(f"PUT {len(items)} attributes to obj_id: {obj_id} bucket: {bucket}")
 
     obj_json = await get_metadata_obj(app, obj_id, bucket=bucket)
-    log.debug(f"PUT attribute obj_id: {obj_id} bucket: {bucket} got json")
-
     if "attributes" not in obj_json:
         log.error(f"unexpected obj data for id: {obj_id}")
         raise HTTPInternalServerError()
 
     attributes = obj_json["attributes"]
-    if attr_name in attributes and not replace:
-        # Attribute already exists, return a 409
-        msg = f"Attempt to overwrite attribute: {attr_name} "
-        msg += f"in obj_id: {obj_id}"
-        log.warn(msg)
-        raise HTTPConflict()
 
-    if replace and attr_name not in attributes:
-        # Replace requires attribute exists
-        msg = f"Attempt to update missing attribute: {attr_name} "
-        msg += f"in obj_id: {obj_id}"
-        log.warn()
-        raise HTTPNotFound()
-
-    if replace:
-        orig_attr = attributes[attr_name]
-        create_time = orig_attr["created"]
-    else:
-        create_time = time.time()
-
-    # ok - all set, create attribute obj
-    attr_json = {"type": datatype,
-                 "shape": shape,
-                 "value": value,
-                 "created": create_time}
-    attributes[attr_name] = attr_json
+    # check for conflicts, also set timestamp
+    create_time = time.time()
+    new_attribute = False  # set this if we have any new attributes
+    for attr_name in items:
+        attribute = items[attr_name]
+        if attr_name in attributes:
+            log.debug(f"attribute {attr_name} exists")
+            if replace:
+                # don't change the create timestamp
+                log.debug(f"attribute {attr_name} exists, but will be updated")
+                old_item = attributes[attr_name]
+                attribute["created"] = old_item["created"]
+            else:
+                # Attribute already exists, return a 409
+                msg = f"Attempt to overwrite attribute: {attr_name} "
+                msg += f"in obj_id: {obj_id}"
+                log.warn(msg)
+                raise HTTPConflict()
+        else:
+            # set the timestamp
+            log.debug(f"new attribute {attr_name}")
+            attribute["created"] = create_time
+            new_attribute = True
+      
+    # ok - all set, create the attributes
+    for attr_name in items:
+        log.debug(f"adding attribute {attr_name}")
+        attr_json = items[attr_name]
+        attributes[attr_name] = attr_json
 
     # write back to S3, save to metadata cache
     await save_metadata_obj(app, obj_id, obj_json, bucket=bucket)
 
-    resp_json = {}
+    if new_attribute:
+        status = 201
+    else:
+        status = 200
 
-    resp = json_response(resp_json, status=201)
+    resp_json = {"status": status}
+
+    resp = json_response(resp_json, status=status)
     log.response(request, resp=resp)
     return resp
-
 
 async def DELETE_Attribute(request):
     """HTTP DELETE method for /(obj)/<id>/attributes/<name>
--- a/hsds/attr_dn.py
+++ b/hsds/attr_dn.py
@@ -137,6 +137,70 @@ async def GET_Attributes(request):
     return resp
 
 
+async def POST_Attributes(request):
+    """ Return JSON for attribute collection
+    """
+    log.request(request)
+    app = request.app
+
+    if not request.has_body:
+        msg = "POST_Attributes with no body"
+        log.warn(msg)
+        raise HTTPBadRequest(reason=msg)
+
+    body = await request.json()
+    if "attributes" not in body:
+        msg = f"POST_Attributes expected attributes in body but got: {body.keys()}"
+        log.warn(msg)
+        raise HTTPBadRequest(reason=msg)
+
+    titles = body["attributes"]  # list of attribute names to fetch
+
+    params = request.rel_url.query
+
+    obj_id = get_obj_id(request)
+    if "bucket" in params:
+        bucket = params["bucket"]
+    else:
+        bucket = None
+
+    include_data = False
+    log.debug(f"got params: {params}")
+    if "IncludeData" in params and params["IncludeData"]:
+        include_data = True
+        log.debug("include attr data")
+
+    obj_json = await get_metadata_obj(app, obj_id, bucket=bucket)
+
+    log.debug(f"Get attributes obj_id: {obj_id} got json")
+    if "attributes" not in obj_json:
+        msg = f"unexpected data for obj id: {obj_id}"
+        msg.error(msg)
+        raise HTTPInternalServerError()
+
+    # return a list of attributes based on sorted dictionary keys
+    attr_dict = obj_json["attributes"]
+    attr_list = []
+
+    for attr_name in titles:
+        if attr_name not in attr_dict:
+            continue
+        src_attr = attr_dict[attr_name]
+        des_attr = {}
+        des_attr["created"] = src_attr["created"]
+        des_attr["type"] = src_attr["type"]
+        des_attr["shape"] = src_attr["shape"]
+        des_attr["name"] = attr_name
+        if include_data:
+            des_attr["value"] = src_attr["value"]
+        attr_list.append(des_attr)
+
+    resp_json = {"attributes": attr_list}
+    resp = json_response(resp_json)
+    log.response(request, resp=resp)
+    return resp
+
+
 async def GET_Attribute(request):
     """HTTP GET method to return JSON for /(obj)/<id>/attributes/<name>
     """
--- a/hsds/attr_dn.py
+++ b/hsds/attr_dn.py
@@ -461,10 +461,10 @@ async def DELETE_Attributes(request):
     else:
         encoding = None
 
-    if "seperator" in params:
-        seperator = params["seperator"]
+    if "separator" in params:
+        separator = params["separator"]
     else:
-        seperator = "/"
+        separator = "/"
 
     if "attr_names" not in params:
         msg = "expected attr_names for DELETE attributes"
@@ -475,7 +475,7 @@ async def DELETE_Attributes(request):
     if encoding:
         attr_names_param = decodeData(attr_names_param).decode("utf-8")
 
-    attr_names = attr_names_param.split(seperator)
+    attr_names = attr_names_param.split(separator)
 
     log.info(f"DELETE attribute {attr_names} in {obj_id} bucket: {bucket}")
 
--- a/hsds/attr_dn.py
+++ b/hsds/attr_dn.py
@@ -196,6 +196,7 @@ async def POST_Attributes(request):
         attr_list.append(des_attr)
 
     resp_json = {"attributes": attr_list}
+    log.debug(f"POST attributes returning: {resp_json}")
     resp = json_response(resp_json)
     log.response(request, resp=resp)
     return resp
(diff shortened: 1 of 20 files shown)


hsds/attr_dn.py
```diff
@@ -203,69 +265,90 @@ async def PUT_Attribute(request):
...
-                 "shape": shape,
-                 "value": value,
-                 "created": create_time}
-    attributes[attr_name] = attr_json
+    # check for conflicts, also set timestamp
+    create_time = time.time()
+    new_attribute = False  # set this if we have any new attributes
+    for attr_name in items:
+        attribute = items[attr_name]
+        if attr_name in attributes:
```
mattjala, line 313: `attributes` should have a name that makes it clear that it's a list of attributes that already exist/are already in storage. Maybe `retrieved_attributes`? This would also be good for `attr_dict` in `POST_Attributes`

hsds/attr_sn.py
```diff
@@ -384,28 +337,318 @@ async def PUT_Attribute(request):
...
+
+    log.debug(f"obj_ids: {obj_ids}")
+
+    # TBD - verify that the obj_id belongs to the given domain
+    await validateAction(app, domain, req_obj_id, username, "create")
+
+    count = len(obj_ids)
+    if count == 0:
+        msg = "no obj_ids defined"
+        log.warn(f"PUT_Attributes: {msg}")
```
mattjala, line 604: This seems to indicate that `PUT_Attributes` is plural with respect to acting on multiple attribute parent objects, but the `PUT_Attributes` in the data node acted on only a single attribute parent object. 

Is this comment saying that, if `count == 1`, the `PUT_Attributes` in the service node acts as the singular `PUT_Attribute`? That doesn't seem accurate, since it could write multiple attributes to a single attribute parent object in one invocation.

hsds/attr_sn.py
```diff
@@ -764,3 +1008,217 @@ async def PUT_AttributeValue(request):
     resp = await jsonResponse(request, req_rsp)
     log.response(request, resp=resp)
     return resp
+
+
+async def _get_attributes(app, obj_id, attr_names,
+                          IncludeData=False,
```
mattjala, line 1014: `IncludeData -> includeData`

hsds/attr_sn.py
```diff
@@ -764,3 +1008,217 @@ async def PUT_AttributeValue(request):
...
+    if ignore_nan:
+        params["ignore_nan"] = 1
+    if bucket:
+        params["bucket"] = bucket
+    data = {"attributes": attr_names}
+    log.debug(f"using params: {params}")
+    dn_json = await http_post(app, req, data=data, params=params)
+    log.debug(f"got attributes json from dn for obj_id: {dn_json}")
+
+    if len(dn_json) < len(attr_names):
```
mattjala, line 1035: Should this check be reworked to `if len(dn_json['attributes']) != len(attr_names)`, and placed after the check that `'attributes'`is in `dn_json`?

hsds/domain_crawl.py
```diff
@@ -0,0 +1,275 @@
...
+        for obj_id in objs:
+            log.debug(f"adding {obj_id} to the queue")
+            self._q.put_nowait(obj_id)
+        if isinstance(objs, dict):
+            self._objs = objs
+        else:
+            self._objs = None
+
+
+    async def get_attributes(self, obj_id, attr_names):
```
mattjala, line 61: Could this be merged with `_get_attributes` from `attr_sn`? The domain crawler `self` seems like it contains the `app` and parameter fields needed for `_get_attributes`.

hsds/domain_crawl.py
```diff
@@ -0,0 +1,275 @@
...
+        # get the given attributes for the obj_id
```
mattjala, line 62: `get the given attributes for the obj_id` -> `write the given attributes to the obj_id`

hsds/domain_crawl.py
```diff
@@ -0,0 +1,275 @@
...
+                msg += f"but only {len(attributes)} were returned"
+                log.warn(msg)
+            self._obj_dict[obj_id] = attributes
+        else:
+            log.warn(f"Domain crawler - got {status} status for obj_id {obj_id}")
+            self._obj_dict[obj_id] = {"status": status}
+       
+
+    async def put_attributes(self, obj_id, attr_items):
+        # get the given attributes for the obj_id
```
mattjala, line 115: `get the given attributes for the obj_id` -> `write the given attributes to the obj_id`

hsds/domain_crawl.py
```diff
@@ -0,0 +1,275 @@
...
+                msg += f"failed, status: {status}"
+                log.error(msg)
+            else:
+                msg += f"returned status: {status}"
+                log.warn(msg)
+            return
+        
+        log.debug(f"DomainCrawler - got json for {obj_id}")
+
+        # if including links, we need link count
```
mattjala, line 181: Should be `we don't need link count`?

tests/integ/attr_test.py
```diff
@@ -1454,6 +1465,381 @@ def testNaNAttributeValue(self):
...
+            self.assertEqual(len(attr_value), extent)
+            self.assertEqual(attr_value, [i*10+j for j in range(extent)])
+
+        # try writing again, should get 409
+        rsp = self.session.put(req, data=json.dumps(data), headers=headers)
+        self.assertEqual(rsp.status_code, 409)
+        """
+
+        # write attributes to the three group objects
+        print(grp_ids)
```
mattjala, line 1809: Print should be removed

tests/integ/attr_test.py
```diff
@@ -1454,6 +1465,381 @@ def testNaNAttributeValue(self):
...
+        attr_count = 4
+        attributes = {}
+        extent = 10
+        for i in range(attr_count):
+            value = [i*10 + j for j in range(extent)]
+            data = {"type": "H5T_STD_I32LE", "shape": extent, "value": value}
+            attr_name = f"attr{i+1:04d}"
+            attributes[attr_name] = data
+        """
+        # write attributes to the dataset
```
mattjala, line 1780: Should be removed or enable

hsds/attr_sn.py
```diff
@@ -384,28 +337,318 @@ async def PUT_Attribute(request):
...
+        raise HTTPBadRequest(reason=msg)
+    elif count == 1:
+        # just send one PUT Attributes request to the dn
+        obj_id = list(obj_ids.keys())[0]
+        attr_json = obj_ids[obj_id]
+        log.debug(f"got attr_json: {attr_json}")
+        req = getDataNodeUrl(app, obj_id)
+        collection = getCollectionForId(obj_id)
+        req += f"/{collection}/{obj_id}/attributes"
+        log.info(f"PUT Attribute: {req}")
```
mattjala, line 614: `PUT Attribute` -> `PUT Attributes`

hsds/attr_sn.py
```diff
@@ -384,28 +337,318 @@ async def PUT_Attribute(request):
...
+        data = {"attributes": attr_json}
+        put_rsp = await http_put(app, req, data=data, params=params)
+        log.info(f"PUT Attribute resp: {put_rsp}")
```
mattjala, line 617: `PUT Attribute -> PUT Attributes`

mattjala:
Here is a list of cases where either HSDS or the client can attempt to use an overlong URL due to attributes or links having long names:

Client request is safe, but can cause HSDS to use an overlong URL
- `POST_Dataset`, dset_sn.py:1182 -> `PUT_Link`
- `POST_Datatype`, ctype_sn.py:238 -> `PUT_Link`
- `POST_Group`, group_sn.py:238 -> `PUT_Link`

Both client request and resultant HSDS request can use an overlong URL
- `PUT_Link`, link_sn.py:286 -> `PUT_Link`
- `GET_Link`, link_sn.py:158 -> `GET_Link`
- `DELETE_Link`, link_sn.py:363 -> `DELETE_Link`
- `GET_Attribute`, attr_sn.py:183 -> `GET_Attribute`
- `GET_AttributeValue`, attr_sn.py:514, -> `GET_Attribute`
- `PUT_Attribute`, attr_sn.py:390 -> `PUT_Attribute`
- `DELETE_Attribute`, attr_sn.py:451, -> `DELETE_Attribute`
- `PUT_AttributeValue`, attr_sn.py:642, attr_sn.py:750 -> `GET_Attribute`, `PUT_Attribute`

These API endpoints expect a potentially overlong `h5path` query string in the URL, and then make a potentially overlong `GET_Link` request in `getObjectIdByPath`
- `GET_Domain`
- `GET_Datatype`
- `GET_Group`
- `GET_Dataset`

`POST_Attributes` and `PUT_Attributes` as implemented in this PR would provide a safe alternative to the `GET_Attribute`/`GET_AttributeValue`/`PUT_Attribute`/`PUT_AttributeValue` API endpoints by storing long attribute names in the request body.

The HDF5 Library doesn't formally specify an attribute name limit, and only attribute names up to 255 bytes are tested. For this reason, I think we can safely consider attribute names that cause a URL to exceed aiohttp's default limit (8190 bytes) to be unsupported. In that case, we wouldn't need to worry about changing `DELETE_Attribute`. 

The changes that would need to be made to avoid all other potentially overlong URLs:
- `PUT_Link` should take link name in the request body instead of the query string.
- `DELETE_Link` would need to be implemented under POST_Link or PUT_Link.
- `GET_*` would need to be implemented under `POST_*` or `PUT_*` for links, groups, datasets, datatypes, and domains.

The latter two changes don't seem like perfect ideas - moving all the GET functionality under POST/PUT could be be unintuitive or duplicate behavior. Some other possible solutions to handle the long URL cases in `GET_*`/`DELETE_*` endpoints:

- Increase the max URL length accepted by HSDS through `aiohttp`, and hope that intermediaries between HSDS and the client accept a URL up to ~67 Kb.
- Do not support link names above ~2048 bytes (generally considered an upper bound on 'reasonable' URL lengths). For reference, the library supports link names up to 65 Kb.
- Allow GET and DELETE requests to provide a body containing the target link name/target h5path. This is probably bad practice, and also requires the user to change how they provide the link name based on how long it is. If we require the link name to always be provided in the body, then that's a breaking API change, but the API going forward would be more consistent.

hsds/attr_dn.py
jreadey: The attributes are the current set of attributes, so I think I'll keep it the way it is.  Realize there's potential for confusion in other methods - will review.

hsds/attr_sn.py
```diff
@@ -384,28 +337,318 @@ async def PUT_Attribute(request):
...
+        log.debug(f"PUT Attribute {len(attr_items)} attibutes to add")
+    else:
+        log.debug("no attributes defined yet")
+
+    # next, sort out where these attributes are going to
+    
+    obj_ids = {}
+    if "obj_ids" in body:
+        body_ids = body["obj_ids"]
+        if isinstance(body_ids, list):
```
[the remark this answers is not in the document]
jreadey: expanded comments to explain this better

hsds/attr_sn.py
jreadey: changed the calling code (wanted to keep the named argument same as the http query param

hsds/attr_sn.py
jreadey: fixed

hsds/domain_crawl.py
jreadey: Good idea. Created a getAttributes function in servicenode_lib.py that both attr_sn and DomainCrawl call.

hsds/domain_crawl.py
jreadey: fixed

hsds/domain_crawl.py
jreadey: Thinking was you don't need it since you can jsut do len(links), but added back to have fewer special cases.

tests/integ/attr_test.py
jreadey: fixed

tests/integ/attr_test.py
jreadey: enabled

hsds/attr_sn.py
jreadey: fixed

hsds/attr_sn.py
jreadey: fixed

tests/integ/attr_test.py
```diff
@@ -1703,7 +1704,14 @@ def testPostAttributeMultiple(self):
         self.assertEqual(len(root_attrs), 2)
         dset_attrs = attributes[dset_id]
         self.assertEqual(len(dset_attrs), 1)  # one of the ones we asked for didn't exist
-    
+
+        # test with no providing all attribute names - should return all attributes
```
mattjala, line 1708: Should be enabled or removed

tests/integ/attr_test.py
```diff
@@ -1798,22 +1806,42 @@ def testPutAttributeMultiple(self):
...
+        data = {"obj_ids": attributes}  # no "attributes" key this time
+        req = self.endpoint + "/groups/" + root_id + "/attributes"
+        rsp = self.session.put(req, data=json.dumps(data), headers=headers)
+        self.assertEqual(rsp.status_code, 201)
+
         # do a get attributes on the three group objects to verify
-        for grp_id in grp_ids:
+        for i in range(grp_count):
+            grp_id = grp_ids[i]
+            print(f"grp_id: {grp_id}")
```
mattjala, line 1844: Remove print statement

mattjala (dismissed):

tests/integ/attr_test.py
```diff
@@ -249,9 +248,8 @@ def testObjAttr(self):
             # try adding the attribute as a different user
             user2_name = config.get("user2_name")
             if user2_name:
-                headers = helper.getRequestHeaders(
-                    domain=self.base_domain, username="test_user2"
-                )
+                headers = helper.getRequestHeaders(domain=self.base_domain, username="test_user2")
+                attr_payload = {"type": "H5T_STD_I32LE", "value": 42}
```
mattjala, line 252: This test fails due to `attr_payload` being undefined if `user2_name` isn't found in the config - this should either have a default second username, or the requests should only be made if a second username is found.

hsds/attr_dn.py
```diff
@@ -361,15 +448,40 @@ async def DELETE_Attribute(request):
...
+    if "encoding" in params:
+        encoding = params["encoding"]
+        if encoding != "base64":
+            msg = "only base64 encoding is supported"
+            log.warn(msg)
+            raise HTTPBadRequest(reason=msg)
+    else:
+        encoding = None
+
+    if "seperator" in params:
```
mattjala, line 464: `seperator` -> `separator`

hsds/attr_sn.py
```diff
@@ -1195,3 +1316,84 @@ async def POST_Attributes(request):
...
+    if "encoding" in params:
+        encoding = params["encoding"]
+        if encoding != "base64":
+            msg = "only base64 encoding is supported for attribute names"
+            log.warn(msg)
+            raise HTTPBadRequest(reason=msg)
+    else:
+        encoding = None
+
+    if "seperator" in params:
```
mattjala, line 1355: `seperator` -> `separator`

hsds/servicenode_lib.py
```diff
@@ -539,3 +642,136 @@ async def doFlush(app, root_id, bucket=None):
...
+    else:
+        status = 201
+
+    log.info(f"putAttributes status: {status}")
+
+    return status
+
+
+async def deleteAttributes(app, obj_id, attr_names=None, seperator="/", bucket=None):
+    """ get the requested set of attributes from the given object """
```
mattjala, line 756: Description should be `delete the requested set of attributes from the given object`

tests/integ/attr_test.py
```diff
@@ -1454,6 +1588,526 @@ def testNaNAttributeValue(self):
...
+
+        # try deleting the attribute by name
+        rsp = self.session.delete(bad_req, headers=headers)
+        self.assertEqual(rsp.status_code, 404)  # not found
+
+        # send attribute name as an encoded query param
+        attr_names_param = base64.b64encode(attr_name.encode("utf8")).decode("ascii")
+        # specify a seperator since our attribute name has the default slash
+        params = {"attr_names": attr_names_param, "encoding": "base64", "seperator": "!"}
+        rsp = self.session.delete(req, params=params, headers=headers)
```
mattjala, line 1691: I think we should include a test that deletes multiple attributes at once through `DELETE_Attributes`.  I wrote the following while testing the API:

```
    def testDeleteAttributesMultiple(self):
        print("testDeleteAttributesMultiple", self.base_domain)
        headers = helper.getRequestHeaders(domain=self.base_domain)
        req = self.endpoint + "/"

        # Get root uuid
        rsp = self.session.get(req, headers=headers)
        rspJson = json.loads(rsp.text)
        root_uuid = rspJson["root"]
        helper.validateId(root_uuid)

        # create a subgroup
        req = self.endpoint + "/groups"
        rsp = self.session.post(req, headers=headers)
        rspJson = json.loads(rsp.text)
        grp_id = rspJson["id"]
        self.assertTrue(helper.validateId(grp_id))

        # link as "grp1"
        grp_name = "grp1"
        req = self.endpoint + "/groups/" + root_uuid + "/links/" + grp_name
        payload = {"id": grp_id}
        rsp = self.session.put(req, data=json.dumps(payload), headers=headers)
        self.assertEqual(rsp.status_code, 201)  # created

        # Create attributes
        for i in range(10):
            attr_name = "attr" + str(i * 100)
            value = [i]
            data = {"type": "H5T_IEEE_F32LE", "shape": 1, "value": value}
            req = self.endpoint + "/groups/" + grp_id + "/attributes/" + attr_name
            rsp = self.session.put(req, data=json.dumps(data), headers=headers)
            self.assertEqual(rsp.status_code, 201) # Created

        # Delete all by parameter
        attr_names = ["attr" + str(i * 100) for i in range(10)]
        separator = '/'

        req = self.endpoint + "/groups/" + grp_id + "/attributes?attr_names=" + separator.join(attr_names)
        rsp = self.session.delete(req, headers=headers)

        # Attempt to read deleted attributes
        for i in range(10):
            req = self.endpoint + "/groups/" + grp_id + "/attributes/" + "attr" + str(i)
            rsp = self.session.get(req, headers=headers)
            self.assertEqual(rsp.status_code, 404)

        # Create another batch of attributes
        for i in range(10):
            attr_name = "attr" + str(i * 100)
            value = [i]
            data = {"type": "H5T_IEEE_F32LE", "shape": 1, "value": value}
            req = self.endpoint + "/groups/" + grp_id + "/attributes/" + attr_name
            rsp = self.session.put(req, data=json.dumps(data), headers=headers)
            self.assertEqual(rsp.status_code, 201) # Created

        # Delete with custom separator
        attr_names = ["attr" + str(i * 100) for i in range(10)]
        separator = ':'

        req = self.endpoint + "/groups/" + grp_id + "/attributes?attr_names=" + separator.join(attr_names) + "&seperator=" + separator
        rsp = self.session.delete(req, headers=headers)

        # Attempt to read
        for i in range(10):
            req = self.endpoint + "/groups/" + grp_id + "/attributes/" + "attr" + str(i)
            rsp = self.session.get(req, headers=headers)
            self.assertEqual(rsp.status_code, 404)
```

mattjala:
At this point, there are four different ways to "get the value of an attribute" - `GET_Attribute`, `GET_Attributes`, `GET_AttributeValue`, and `POST_Attributes`, and two different ways to get the values of multiple attributes. Is there a motive for each of these different methods to continue to exist beyond backwards compatibility?

tests/integ/attr_test.py
jreadey: Fixed

hsds/attr_dn.py
jreadey: Fixed

hsds/attr_sn.py
jreadey: Fixed

hsds/servicenode_lib.py
jreadey: Fixed

tests/integ/attr_test.py
jreadey: Good idea. Added test

PR #296, opened 2024-01-08
Multilink

jreadey:
Added post method for links and multi support for delete link

### DIFF
--- a/hsds/domain_sn.py
+++ b/hsds/domain_sn.py
@@ -23,9 +23,8 @@ from aiohttp.web_exceptions import HTTPGone, HTTPInternalServerError
 from aiohttp.web_exceptions import HTTPConflict, HTTPServiceUnavailable
 from aiohttp import ClientResponseError
 from aiohttp.web import json_response
-from requests.sessions import merge_setting
 
-from .util.httpUtil import getObjectClass, http_post, http_put, http_get, http_delete
+from .util.httpUtil import getObjectClass, http_post, http_put, http_delete
 from .util.httpUtil import getHref, respJsonAssemble
 from .util.httpUtil import jsonResponse
 from .util.idUtil import getDataNodeUrl, createObjId, getCollectionForId
@@ -47,60 +46,54 @@ from . import hsds_logger as log
 from . import config
 
 
-async def get_collections(app, root_id, bucket=None):
+async def get_collections(app, root_id, bucket=None, max_objects_limit=None):
     """Return the object ids for given root."""
 
     log.info(f"get_collections for {root_id}")
-    groups = {}
-    datasets = {}
-    datatypes = {}
-    lookup_ids = set()
-    lookup_ids.add(root_id)
-    params = {"bucket": bucket}
 
-    while lookup_ids:
-        grp_id = lookup_ids.pop()
-        req = getDataNodeUrl(app, grp_id)
-        req += "/groups/" + grp_id + "/links"
-        log.debug("collection get LINKS: " + req)
-        try:
-            # throws 404 if doesn't exist
-            links_json = await http_get(app, req, params=params)
-        except HTTPNotFound:
-            log.warn(f"get_collection, group {grp_id} not found")
-            continue
+    crawler_params = {
+        "include_attrs": False,
+        "include_links": False,
+        "bucket": bucket,
+        "follow_links": True,
+    }
 
-        log.debug(f"got links json from dn for group_id: {grp_id}")
-        links = links_json["links"]
-        log.debug(f"get_collection: got links: {links}")
-        for link in links:
-            if link["class"] != "H5L_TYPE_HARD":
-                continue
-            link_id = link["id"]
-            obj_type = getCollectionForId(link_id)
-            if obj_type == "groups":
-                if link_id in groups:
-                    continue  # been here before
-                groups[link_id] = {}
-                lookup_ids.add(link_id)
-            elif obj_type == "datasets":
-                if link_id in datasets:
-                    continue
-                datasets[link_id] = {}
-            elif obj_type == "datatypes":
-                if link_id in datatypes:
-                    continue
-                datatypes[link_id] = {}
-            else:
-                msg = "get_collection: unexpected link object type: "
-                msg += f"{obj_type}"
-                log.error(merge_setting)
-                HTTPInternalServerError()
+    if max_objects_limit:
+        crawler_params["max_objects_limit"] = max_objects_limit
+
+    crawler = DomainCrawler(app, [root_id, ], action="get_obj", params=crawler_params)
+    await crawler.crawl()
+    if max_objects_limit and len(crawler._obj_dict) >= max_objects_limit:
+        msg = "get_collections - too many objects:  "
+        msg += f"{len(crawler._obj_dict)}, returning None"
+        log.info(msg)
+        return None
+    else:
+        msg = f"DomainCrawler returned: {len(crawler._obj_dict)} object ids"
+        log.info(msg)
+
+    group_ids = set()
+    dataset_ids = set()
+    datatype_ids = set()
+
+    for obj_id in crawler._obj_dict:
+        obj_type = getCollectionForId(obj_id)
+        if obj_type == "groups":
+            group_ids.add(obj_id)
+        elif obj_type == "datasets":
+            dataset_ids.add(obj_id)
+        elif obj_type == "datatypes":
+            datatype_ids.add(obj_id)
+        else:
+            log.warn(f"get_collections - unexpected id type: {obj_id}")
+    if root_id in group_ids:
+        group_ids.remove(root_id)  # don't include the root id
+    print(f"get_collections - group_ids: {group_ids}")
 
     result = {}
-    result["groups"] = groups
-    result["datasets"] = datasets
-    result["datatypes"] = datatypes
+    result["groups"] = group_ids
+    result["datasets"] = dataset_ids
+    result["datatypes"] = datatype_ids
     return result
 
 
@@ -114,9 +107,10 @@ async def getDomainObjects(app, root_id, include_attrs=False, bucket=None):
 
     crawler_params = {
         "include_attrs": include_attrs,
-        "bucket": bucket,
+        "include_links": True,
         "follow_links": True,
         "max_objects_limit": max_objects_limit,
+        "bucket": bucket,
     }
 
     crawler = DomainCrawler(app, [root_id, ], action="get_obj", params=crawler_params)
@@ -263,15 +257,13 @@ async def get_domains(request):
             if pattern:
                 # do a pattern match on the basename
                 basename = op.basename(domain)
-                log.debug(
-                    f"get_domains: checking {basename} against pattern: {pattern}"
-                )
+                msg = f"get_domains: checking {basename} against pattern: {pattern}"
+                log.debug(msg)
                 try:
                     got_match = globmatch(basename, pattern)
                 except ValueError as ve:
-                    log.warn(
-                        f"get_domains, invalid query pattern {pattern}, ValueError: {ve}"
-                    )
+                    msg = f"get_domains, invalid query pattern {pattern}, ValueError: {ve}"
+                    log.warn(msg)
                     raise HTTPBadRequest(reason="invalid query pattern")
                 if got_match:
                     log.debug("get_domains - got_match")
@@ -502,14 +494,14 @@ async def GET_Domain(request):
         h5path = params["h5path"]
 
         # select which object to perform path search under
-        root_id = parent_id if parent_id else domain_json["root"]
+        base_id = parent_id if parent_id else domain_json["root"]
 
         # getObjectIdByPath throws 404 if not found
         obj_id, domain, _ = await getObjectIdByPath(
-            app, root_id, h5path, bucket=bucket, domain=domain,
+            app, base_id, h5path, bucket=bucket, domain=domain,
             follow_soft_links=follow_soft_links,
             follow_external_links=follow_external_links)
-        log.info(f"get obj_id: {obj_id} from h5path: {h5path}")
+        log.info(f"got obj_id: {obj_id} from h5path: {h5path}")
         # get authoritative state for object from DN (even if
         # it's in the meta_cache).
         kwargs = {"refresh": True, "bucket": bucket,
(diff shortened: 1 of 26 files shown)


hsds/link_dn.py
```diff
@@ -136,86 +136,146 @@ async def GET_Links(request):
     return resp
 
 
-async def GET_Link(request):
-    """HTTP GET method to return JSON for a link"""
+async def POST_Links(request):
+    """HTTP POST method to return JSON for a link a given set of links """
```
mattjala, line 140: `for a link a given set of links` -> `for a link or a given set of links`

hsds/link_dn.py
```diff
@@ -136,86 +136,146 @@ async def GET_Links(request):
...
+        else:
+            log.warn(f"unexpected to link class {link_class} for link: {title}")
+            continue
+
+        item["title"] = title
+
+        link_list.append(item)
+
+    if not link_list:
+        msg = f"POST link - requested {len(titles)} but none were found"
```
mattjala, line 227: `POST link` -> `POST_Links`

hsds/link_sn.py
```diff
@@ -236,99 +230,232 @@ async def PUT_Link(request):
...
     # returns 201 if new link was created, 200 if this is a duplicate
     # of an existing link
-    resp = await jsonResponse(request, req_rsp, status=dn_status)
+    resp = await jsonResponse(request, req_rsp, status=status)
+    log.response(request, resp=resp)
+    return resp
+
+
+async def DELETE_Links(request):
+    """HTTP method to delete multiple link"""
```
mattjala, line 260: `link` -> `links`

hsds/link_sn.py
```diff
@@ -236,99 +230,232 @@ async def PUT_Link(request):
...
+    elif len(items) == 1 and not follow_links:
+        # just make a request to the datanode
+        group_id = list(items.keys())[0]
+        titles = items[group_id]
+        links = await getLinks(app, group_id, titles=titles, bucket=bucket)
+
+        resp_json["links"] = links
+    else:
+        # get multi obj
+        # don't follow links for the groups we visit!
```
mattjala, line 436: This comment seems misleading, since domain crawlers can follow links based on the parameter in `domain_crawl.py/get_links`.

hsds/servicenode_lib.py
```diff
@@ -341,6 +342,158 @@ async def getDsetJson(app, dset_id,
...
+    """ create a new hard link.  Return 201 if this is a new link,
+      or 201 if it's a duplicate of an existing link """
+
+    status = await putLink(app, group_id, title, tgt_id=tgt_id, bucket=bucket)
+    return status
+
+
+async def putSoftLink(app, group_id, title, h5path=None, bucket=None):
+    """ create a new hard link.  Return 201 if this is a new link,
+      or 201 if it's a duplicate of an existing link """
```
mattjala, line 463: `hard link` -> `soft link`

hsds/servicenode_lib.py
```diff
@@ -341,6 +342,158 @@ async def getDsetJson(app, dset_id,
...
+
+    status = await putLink(app, group_id, title, h5path=h5path, bucket=bucket)
+    return status
+
+
+async def putExternalLink(app, group_id, title, h5path=None, h5domain=None, bucket=None):
+    """ create a new hard link.  Return 201 if this is a new link,
```
mattjala, line 470: `hard link` -> `external link`

hsds/servicenode_lib.py
```diff
@@ -341,6 +342,158 @@ async def getDsetJson(app, dset_id,
...
+      or 201 if it's a duplicate of an existing link """
+
+    status = await putLink(app, group_id, title, h5path=h5path, h5domain=h5domain, bucket=bucket)
+    return status
+
+
+async def deleteLinks(app, group_id, titles=None, separator="/", bucket=None):
+    """ delete the requested set of attributes from the given object """
```
mattjala, line 478: `attributes from the given object` -> `links from the given group`

mattjala:
There's a REST VOL url-encoding test with a group named ``url_encoding_group !*'():@&=+$,?#[]-.<>\\^`{}|~``, and it fails with a 404 when trying to re-open the group in this PR. I'll see if I can track down the exact cause.

hsds/link_dn.py
jreadey: Fixed

hsds/link_dn.py
jreadey: Fixed

hsds/link_sn.py
jreadey: Fixed

hsds/link_sn.py
jreadey: Fixed

hsds/servicenode_lib.py
jreadey: Fixed

hsds/servicenode_lib.py
jreadey: Fixed

jreadey: Fixed

hsds/servicenode_lib.py
jreadey: Fixed

mattjala:
The issue with URL encoding is actually on the REST VOL's side. This PR fixed an issue that was obscuring it.  LGTM for now. Details on the encoding issue are below.

The issue comes from backslashes in the link name. Before this PR, the link name ``url_encoding_group !*'():@&=+$,?#[]-.<>\^`{}|~`` was erroneously truncated before the `?` during both the creation of the link and the GET request to check for its existence, so the issue went unnoticed. 

After this PR, the link is successfully created with the entire name. When creating the link,  the name is provided as in body of a `POST_Group` request, and its value contains two literal backslash characters. When making the subsequent `GET_Link` request, the same link name string is provided as part of the destination URL, and is parsed in `GET_Domain` with four literal backslash characters. 

I think this is because backslashes in the URL don't need to be escaped since they are (or should be) already percent-encoded, but backslashes in JSON strings _do_ need to be escaped with another backslash. So when `GET_Domain` receives two backslashes characters, each is interpreted as a logical backslash character, resulting in four total literal backslashes. The issue is actually on the REST VOL's side, since it should encode strings for JSON properly

mattjala (dismissed):

hsds/util/attrUtil.py
```diff
@@ -51,3 +51,32 @@ def validateAttributeName(name):
...
+        return False
+    shape_class = attr1["shape"].get("class")
+    if shape_class == "H5S_NULL":
+        return True  # nothing else to compare
+    for obj in (attr1, attr2):
+        if "value" not in obj:
+            raise TypeError("expected value key for attribute")
+    return attr1["value"] == attr2["value"]
+
+    if not isinstance(attr1, dict):
```
mattjala, line 78: Unreachable statements

mattjala:
This looks good - I'll approve once I have a build of the REST VOL working with the new responses.

hsds/util/attrUtil.py
jreadey: Removed

hsds/domain_sn.py
```diff
@@ -624,6 +624,131 @@ async def getScanTime(app, root_id, bucket=None):
...
+    app = request.app
+    params = request.rel_url.query
+    log.debug(f"POST_Domain query params: {params}")
+
+    include_links = False
+    include_attrs = False
+    follow_soft_links = False
+    follow_external_links = False
+
+    if "include_links" in params and params["include_links"]:
```
mattjala, line 640: This needs to check for `parent_id` for relative path searches, and use `parent_id` if it exists for `getObjectIdByPath`

hsds/domain_sn.py
```diff
@@ -624,6 +624,131 @@ async def getScanTime(app, root_id, bucket=None):
...
+
+        obj_json = respJsonAssemble(obj_json, params, obj_id)
+
+        obj_json["domain"] = getPathForDomain(domain)
+
+        # client may not know class of object retrieved via path
+        obj_json["class"] = getObjectClass(obj_id)
+
+        json_objs[h5path] = obj_json
+
```
mattjala, line 745: Some response parsing callbacks in the REST VOL use hrefs to get information about the target domain, so `POST_Domain` not returning any hrefs becomes a problem. As a first pass, this solved the problem by attaching hrefs to each returned object individually (though a complete version might need to handle when some objects have different domains due to external links):

```

    hrefs = []
    hrefs.append({"rel": "self", "href": getHref(request, "/")})
    if "root" in domain_json:
        root_uuid = domain_json["root"]
        href = getHref(request, "/datasets")
        hrefs.append({"rel": "database", "href": href})
        href = getHref(request, "/groups")
        hrefs.append({"rel": "groupbase", "href": href})
        href = getHref(request, "/datatypes")
        hrefs.append({"rel": "typebase", "href": href})
        href = getHref(request, "/groups/" + root_uuid)
        hrefs.append({"rel": "root", "href": href})
        href = getHref(request, "/")
        hrefs.append({"rel": "home", "href": href})

    hrefs.append({"rel": "acls", "href": getHref(request, "/acls")})
    parent_domain = getParentDomain(domain)
    if not parent_domain or getPathForDomain(parent_domain) == "/":
        is_toplevel = True
    else:
        is_toplevel = False

    if not is_toplevel:
        href = getHref(request, "/", domain=parent_domain)
        hrefs.append({"rel": "parent", "href": href})

    for h5path in json_objs:
        (json_objs[h5path])["hrefs"] = hrefs
```

mattjala:
Besides my two comments, this seems good to go in - I have a draft PR to the REST VOL which works with a version of this PR that incorporates the changes I requested

hsds/domain_sn.py
jreadey: I've come to think of the hrefs as mainly for human consumption - i.e. clicking on a link that shows up in the browser.  As such, hrefs for POST and PUT don't seem as useful.
Another issue is that some of the POST responses could potentially involve a lot of hrefs in the response - e.g. multiple hrefs for each link in POST_links response.
Could you describe what info that you are currently pulling out of the hrefs?  How about we just include whatever that is in the POST response as regular json?

hsds/domain_sn.py
jreadey: fixed

jreadey:
I've checked in some new code to let you return links based on a regex-like pattern.  Also added a follow_links paran to GET_Links.

hsds/domain_sn.py
mattjala: The REST VOL gets the domain that contains the object from the hrefs, to satisfy some library API calls that retrieve the object's file number (`H5Oget_info3`). It stores the hash of the domain name as a file number, so that comparisons between different objects' file numbers still work.

Returning the Domain that contains the object directly could also solve this. The REST VOL uses this information in `GET_Datatype/Group/Dataset` and `POST_Domain` requests.

tests/integ/link_test.py
```diff
@@ -529,6 +546,133 @@ def testGet(self):
...
+        self.assertTrue("links" in rspJson)
+        grp_links = rspJson["links"]
+        hardlink_count = 0
+        softlink_count = 0
+        extlink_count = 0
+        expected_group_links = ("g1", "g2", "g1.1", "g1.2", "g1.2.1", )
+        expected_dset_links = ("dset1.1.1", "dset1.1.2", "dset2.1", "dset2.2")
+        expected_soft_links = ("slink", )
+        expected_external_links = ("extlink", )
+        self.assertEqual(len(grp_links), 6)
```
mattjala, line 584: Nitpick: It's not clear what this 6 corresponds to, since `expected_group_links` has 5 entries. 
Does `grp_links` refer to 'links under the targeted group in the request' rather than 'links to groups'? Renaming `grp_links` would make this more clear.

tests/integ/link_test.py
```diff
@@ -507,6 +507,23 @@ def testGet(self):
         expected_uuid = helper.getUUIDByPath(domain, "/g1/g1.2/g1.2.1", session=self.session)
         self.assertEqual(expected_uuid, g1_2_1_uuid)
 
+        # do get with a regex pattern
+        params = {"pattern": "ext*"}
```
mattjala, line 511: This section should probably be moved under `testGetPattern`

tests/integ/link_test.py
```diff
@@ -1115,7 +1259,7 @@ def testPostLinkMultiple(self):
         obj_links = rspJson["links"]
         self.assertEqual(len(obj_links), 6)
         expected_group_links = ("g1", "g2", "g1.1", "g1.2", "g1.2.1", )
-        expected_dset_links = ("dset1.2", "dset2.2", "dset1.1.1", "dset1.1.2", "dset2.1", )
```
mattjala: Was there a reason to remove `dset1.2` from this test? It doesn't cause any issues when added back in.

hsds/dset_lib.py
```diff
@@ -904,3 +904,22 @@ async def reduceShape(app, dset_json, shape_update, bucket=None):
...
+
+
+async def deleteAllChunks(app, dset_id, bucket=None):
+    """ Delete any allocated chunks for the given dataset """
+
+    log.info(f"deleteAllChunks for {dset_id}")
+
+    # get all chunk ids for chunks that have been allocated
+    chunk_ids = await getAllocatedChunkIds(app, dset_id, bucket=bucket)
+    chunk_ids.sort()
```
mattjala, line 916: Double sort

hsds/link_dn.py
```diff
@@ -109,6 +114,16 @@ async def GET_Links(request):
         titles.sort()  # sort by key
```
mattjala, line 114: `globmatch` returns the link titles in arbitrary order, so it should probably be moved before this

hsds/link_sn.py
```diff
@@ -76,31 +62,79 @@ async def GET_Links(request):
...
+        msg = f"DomainCrawler returned: {len(crawler._obj_dict)} objects"
+        log.info(msg)
+        links = crawler._obj_dict
+        if pattern:
+            for grp_id in links.keys():
+                grp_links = links[grp_id]
+                ret_links = []
+                for link in grp_links:
+                    title = link["title"]
+                    if globmatch(title, pattern):
```
mattjala, line 105: Is this use of globmatch redundant, since `domain_crawl.sn/get_links` -> `getLinks` ->`GET_Links (DN)` already does the pattern matching?

mattjala (approved):

PR #298, opened 2024-01-24
Fix selection of multiple fields in body

mattjala:
Also adds additional tests for field selections with more than one field.


jreadey:
Lgtm

PR #302, opened 2024-01-29
group post with h5path option

jreadey:
Add h5path option to post groups

### DIFF
--- a/hsds/group_sn.py
+++ b/hsds/group_sn.py
@@ -16,15 +16,15 @@
 from aiohttp.web_exceptions import HTTPBadRequest, HTTPForbidden, HTTPNotFound
 from json import JSONDecodeError
 
-from .util.httpUtil import http_post, getHref
-from .util.httpUtil import jsonResponse
-from .util.idUtil import isValidUuid, getDataNodeUrl, createObjId
+from .util.httpUtil import getHref, jsonResponse, getBooleanParam
+from .util.idUtil import isValidUuid
 from .util.authUtil import getUserPasswordFromRequest, aclCheck
 from .util.authUtil import validateUserPassword
 from .util.domainUtil import getDomainFromRequest, isValidDomain
 from .util.domainUtil import getBucketForDomain, getPathForDomain, verifyRoot
-from .servicenode_lib import getDomainJson, getObjectJson, validateAction, deleteObj
-from .servicenode_lib import getObjectIdByPath, getPathForObjectId, putHardLink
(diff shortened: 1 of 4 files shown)


hsds/servicenode_lib.py
```diff
@@ -1006,3 +1011,116 @@ async def deleteObj(app, obj_id, bucket=None):
...
+
+async def createGroupByPath(app,
+                            parent_id=None,
+                            h5path=None,
+                            implicit=False,
+                            creation_props=None,
+                            bucket=None):
+
+    """ create the group at the designated path relative to the parent.
+    If implicit is True, make any intermidiate groups needed in the h5path. """
```
mattjala, line 1040: `intermidiate` -> `intermediate`

hsds/servicenode_lib.py
```diff
@@ -1006,3 +1011,116 @@ async def deleteObj(app, obj_id, bucket=None):
     meta_cache = app["meta_cache"]
     if obj_id in meta_cache:
         del meta_cache[obj_id]  # remove from cache
+
+
+async def createGroup(app, root_id=None, creation_props=None, bucket=None):
```
mattjala, line 1016: Since the parent group that will contain the new group doesn't have to be the root group, `root_id` should be renamed to `parent_id`

hsds/group_sn.py
```diff
@@ -177,13 +177,16 @@ async def POST_Group(request):
     aclCheck(app, domain_json, "create", username)
 
     verifyRoot(domain_json)
+    root_id = domain_json["root"]
 
-    link_id = None
-    link_title = None
+    # allow parent group creation or not
+    implicit = getBooleanParam(params, "implicit")
```
mattjala, line 183: Whether or not to create intermediate groups is specified in the creation properties of a group - see docs [here](https://docs.hdfgroup.org/hdf5/develop/group___l_c_p_l.html#ga66c4c5d3f34e5cf65d00e47a5387383c). Would it be possible to read this from provided creation properties instead?

hsds/servicenode_lib.py
jreadey: Fixed

hsds/servicenode_lib.py
jreadey: This method is for anonymous group so parent I’d wouldn’t apply.  The root I’d basically determines which domain the group will belong to

jreadey: ID not Id

PR #305, opened 2024-01-31
post with h5path for datatypes and datasets

jreadey:
This commit adds a h5path option for post datasets and post datatypes

### DIFF
--- a/hsds/servicenode_lib.py
+++ b/hsds/servicenode_lib.py
@@ -1024,7 +1024,7 @@ async def createObject(app,
                        layout=None,
                        creation_props=None,
                        bucket=None):
-    """ create a group, ctype, or dataset object and return group json
+    """ create a group, ctype, or dataset object and return object json
         Determination on whether a group, ctype, or dataset is created is based on:
             1) if obj_type and obj_shape are set, a dataset object will be created
             2) if obj_type is set but not obj_shape, a  datatype object will be created
--- a/hsds/servicenode_lib.py
+++ b/hsds/servicenode_lib.py
@@ -242,9 +242,13 @@ async def validateAction(app, domain, obj_id, username, action):
         aclCheck(app, domain_json, action, username)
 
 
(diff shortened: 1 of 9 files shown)


hsds/servicenode_lib.py
```diff
@@ -1013,41 +1017,78 @@ async def deleteObj(app, obj_id, bucket=None):
...
-    log.info(f"new group id: {group_id}")
-    group_json = {"id": group_id, "root": root_id}
+async def createObject(app,
+                       root_id=None,
+                       obj_type=None,
+                       obj_shape=None,
+                       layout=None,
+                       creation_props=None,
+                       bucket=None):
+    """ create a group, ctype, or dataset object and return group json
```
mattjala, line 1027: `return group json` -> `return object's json`

hsds/servicenode_lib.py
```diff
@@ -1013,41 +1017,78 @@ async def deleteObj(app, obj_id, bucket=None):
...
+        Determination on whether a group, ctype, or dataset is created is based on:
+            1) if obj_type and obj_shape are set, a dataset object will be created
+            2) if obj_type is set but not obj_shape, a  datatype object will be created
+            3) otherwise (type and shape are both None), a group object will be created
+        The layout parameter only applies to dataset creation
+    """
+    if obj_type and obj_shape:
```
mattjala, line 1034: Is there reason to determine what kind of object to create in this way, instead of just using `obj_type` to carry that information?

hsds/servicenode_lib.py
jreadey: Fixed

hsds/servicenode_lib.py
jreadey: Obj_type is actually the type json fir a dataset or committed data type not an identifier to distinguish between group dataset and ctype   Guess the name is a bit confusing.

PR #307, opened 2024-02-09
Fix POST_Links incorrectly returning all links

mattjala:
When providing a list of link titles and a list of group ids, `POST_Links` is ignoring the provided titles and returning all links in the provided groups. It should return links with the specified titles from each target group, similar to `PUT_Links`.


jreadey (approved):
Lgtm

PR #311, opened 2024-02-15
Fix link test using old param name

mattjala:
The test used the old parameter `group_ids` instead of the new `grp_ids`, causing the parameter to go unrecognized and the test to fail every time.


jreadey (approved):
Lgtm

PR #313, opened 2024-02-16
Fix POST_Links return all on title search with grp list

### LINKED ISSUE #308: POST_Links does not return link information when following links recursively
When using `POST_Links` to retrieve a link by name with `follow_links` enabled, if the link is retrieved from a sub-group, the returned JSON is empty.

Implementing the `follow_links` parameter for `POST_Links` raises some questions - if multiple links are found by the same name, which should be returned? The first in iteration order? Do the domain crawlers guarantee any particular iteration order, and do we want to expose that behavior?

It might make sense to remove the `follow_links` param from `POST_Links`. Similarly, `CreateOrder`, `Limit`, and `Marker` seem to be holdovers from `GET_Links` and should probably be removed as well.

Test to elicit the bug:

```
    def testPostLinksFollowLinks(self):
        domain = self.base_domain + "/testPostLinksFollowLinks.h5"
        helper.setupDomain(domain)
        print("testPostLinksFollowLinks", domain)
        headers = helper.getRequestHeaders(domain=domain)

        req = helper.getEndpoint() + "/"
        rsp = self.session.get(req, headers=headers)
        rspJson = json.loads(rsp.text)
        root_id = rspJson["root"]

        # create group "g1" in root group
        req = helper.getEndpoint() + "/groups"
        body = {"link": {"id": root_id, "name": "g1"}}
        rsp = self.session.post(req, data=json.dumps(body), headers=headers)
        rspJson = json.loads(rsp.text)
        group_id = rspJson["id"]

        path = "/dummy_target"

        # create link "link1" in g1
        req = helper.getEndpoint() + "/groups/" + group_id + "/links/link1"
        body = {"h5path": path}
        rsp = self.session.put(req, data=json.dumps(body), headers=headers)

        # make POST_Links request to root group, expect 404 
        body = {"titles": ["link1"]}
        req = helper.getEndpoint() + "/groups/" + root_id + "/links"
        rsp = self.session.post(req, data=json.dumps(body), headers=headers)
        self.assertEqual(rsp.status_code, 404)

        # make POST_Links request to root group with follow_links, expect to find link1
        req = helper.getEndpoint() + "/groups/" + root_id + "/links?follow_links=1"
        rsp = self.session.post(req, data=json.dumps(body), headers=headers)
        rspJson = json.loads(rsp.text)
        self.assertTrue("links" in rspJson)
        links = rspJson["links"]
        self.assertTrue(len(links) == 1) # fails due to empty return
        link = links[0]
        self.assertTrue("title" in link)
        self.assertEqual(link["title"] == "link1")
```

mattjala:
Fixes #308

### DIFF
--- a/hsds/domain_crawl.py
+++ b/hsds/domain_crawl.py
@@ -323,32 +323,42 @@ class DomainCrawler:
             log.warn(f"get_links, expected groups id but got: {grp_id}")
             return
         kwargs = {"bucket": self._bucket}
-        if titles:
-            kwargs["titles"] = titles
-        else:
-            # only apply these parameters if we are attempting to fetch all links
+
+        # only apply these parameters if we are attempting to fetch all links
+        if not titles:
             if self._limit:
                 kwargs["limit"] = self._limit
             if self._create_order:
                 kwargs["create_order"] = True
 
-            if self._pattern:
-                if self._follow_links:
-                    # apply the pattern after we get the links back,
-                    # otherwise we won't get the groups links that we
-                    # need to follow
-                    log.debug("will apply pattern on return")
-                    pattern = self._pattern
-                else:
-                    kwargs["pattern"] = self._pattern
+        if self._pattern:
+            if self._follow_links:
+                # apply the pattern after we get the links back,
+                # otherwise we won't get the groups links that we
+                # need to follow
+                log.debug("will apply pattern on return")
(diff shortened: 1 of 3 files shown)


mattjala:
We decided that instead of supporting `follow_links` with `titles`, it makes more sense to disable this combination of parameters on this endpoint - see #314

PR #319, opened 2024-02-28
Restrict bucket names to alphanumeric characters and underscores

mattjala:
This fixes a security issue where bucket strings could end up on the command line when using the Azure backend. 

Not all of the [AWS bucket naming rules](https://docs.aws.amazon.com/AmazonS3/latest/userguide/bucketnamingrules.html) are enforced - only the ones that are relevant to preventing security exploits, and can be checked directly by HSDS.

Also fixed PUT_Attributes and POST_Datatype in the DN not retrieving bucket from the body.


jreadey (commented):
What about HSDS running on Posix or Azure?  The AWS rules maybe too restrictive in this case.

jreadey (approved):
lgtm

PR #322, opened 2024-03-04
Fix docker compose not creating multiple SNs correctly

mattjala:
Setting up HSDS through Docker with more than one SN would end up infinitely waiting for the server to be ready.


mattjala:
The issue appears to have been due to my environment variable setup, not docker-compose - I'm not able to replicate the issue.

PR #325, opened 2024-03-11
Fix writing base64 encoded vlen data to point selection

mattjala:
When base64-encoded variable-length data to a point selection, HSDS converts the received base64 directly to a numpy array of the appropriate vlen type in `PUT_Value -> _getRequestData`.  `write_point_sel`, expects the provided data to be JSON and attempts to parse it to an array again, causing an error.

This changes `write_point_sel` to check if its received data is already decoded.


jreadey (approved):
looks good.

PR #326, opened 2024-03-15
Add windows build and test to github workflows

mattjala:
The stdout reconfigure is necessary because python attempts to use the windows 1252 encoding by default on (some versions of) windows, which gets an encoding error in the UTF-8 specific tests.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -14,7 +14,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest]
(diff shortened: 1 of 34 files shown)


jreadey (commented):
When does the sys.stdout.reconfigure call get executed?  Seems weird to be just dangling in this file.  Does it absolutely need to run before any other code?
Otherwise, you can drop it into: basenode.py - baseInit().

mattjala:
- Added workflows to test h5pyd.

The failures when testing h5pyd are due to issue HDFGroup/h5pyd#175 on Windows.

jreadey (approved):
looks good

Issue #332, opened 2024-03-26
not enough room in chunk cache - return 503


assaron:
Hi, I continue to try upload the large file from https://github.com/HDFGroup/hsds/issues/329 and I still get the issues with the upload failing because of too many 503 errors. 

Currently I have a situation when the upload fails very fast after starting, I guess because the server got into some bad state. Here is the relevant portion of hsload log:
```
Mar 26 18:53:41 got link: submission_date
Mar 26 18:53:41 got link: taxid_ch1
Mar 26 18:53:41 got link: title
Mar 26 18:53:41 got link: type
Mar 26 18:53:41 write_dataset src: /data/expression to tgt: /data/expression, shape: (53511, 932405), type: uint32
Mar 26 18:53:41 iterating over chunks for /data/expression
Mar 26 18:53:41 urllib3.exceptions.ResponseError: too many 500 error responses
Mar 26 18:53:41 
Mar 26 18:53:41 The above exception was the direct cause of the following exception:
Mar 26 18:53:41 
Mar 26 18:53:41 Traceback (most recent call last):
Mar 26 18:53:41   File "/home/alserg/.local/pipx/venvs/h5pyd/lib/python3.10/site-packages/requests/adapters.py", line 486, in send
Mar 26 18:53:41     resp = conn.urlopen(
Mar 26 18:53:41   File "/home/alserg/.local/pipx/venvs/h5pyd/lib/python3.10/site-packages/urllib3/connectionpool.py", line 948, in urlopen
Mar 26 18:53:41     return self.urlopen(
Mar 26 18:53:41   File "/home/alserg/.local/pipx/venvs/h5pyd/lib/python3.10/site-packages/urllib3/connectionpool.py", line 948, in urlopen
Mar 26 18:53:41     return self.urlopen(
Mar 26 18:53:41   File "/home/alserg/.local/pipx/venvs/h5pyd/lib/python3.10/site-packages/urllib3/connectionpool.py", line 948, in urlopen
Mar 26 18:53:41     return self.urlopen(
Mar 26 18:53:41   File "/home/alserg/.local/pipx/venvs/h5pyd/lib/python3.10/site-packages/urllib3/connectionpool.py", line 938, in urlopen
Mar 26 18:53:41     retries = retries.increment(method, url, response=response, _pool=self)
Mar 26 18:53:41   File "/home/alserg/.local/pipx/venvs/h5pyd/lib/python3.10/site-packages/urllib3/util/retry.py", line 515, in increment
[...]

PR #333, opened 2024-03-28
Listkeys

jreadey:
Break up list_keys in fileClient with some asyncio.sleep()'s so that other tasks have a chance to run (c.f. issue #332).
Also, reduce log verbosity.


assaron:
@jreadey hope I'm not invading here, but did you mean the opposite condition here: https://github.com/HDFGroup/hsds/blob/c87fa68392644f5530761ceab1a855782e87b3f0/hsds/util/fileClient.py#L356 ? As I understand, currently it sleeps after each file, except for 1000th, but I guess the idea was to sleep only each 1000th file.

mattjala:
I second @assaron's observation. Other than that, LGTM.

jreadey:
Yes, silly mistake!

jreadey:
I'm consistent at least...

PR #335, opened 2024-03-29
Allow periods and hyphens in bucket names

mattjala:
Some NREL buckets have hyphens, so we should definitely allow those. Allowing periods make HSDS at least as permissive as AWS's bucket name rules.


jreadey (approved):
lgtm

PR #338, opened 2024-04-04
update for multi-storage drivers

jreadey:
This PR enables HSDS to support different storage back ends (posix, s3, azure blob) in the same running instance.
If the bucket query parameter is pre-appended with a storage specifier, the corresponding driver will be used for reads and writes.
Currently supported prefixes are:

- "s3://" - aws s3
- "file://" - posix 
- "https://account_name>.blob.windows.net/" - azure  blob storage

If no prefix is used, the default driver will be used.  Where default is the first one meeting the given precondition from this list:

- "file://" - if root_dir config is set
- "s3://" - if aws_s3_gateway is set
- "https....net/" - if azure_connection_string is set
- "file://" - if none of the above apply

### DIFF
--- a/hsds/datanode_lib.py
+++ b/hsds/datanode_lib.py
@@ -327,6 +327,7 @@ async def get_metadata_obj(app, obj_id, bucket=None):
     log.info(f"get_metadata_obj: {obj_id} bucket: {bucket}")
     if isValidDomain(obj_id):
         domain_bucket = getBucketForDomain(obj_id)
+        log.debug(f"tbd - domain_bucket: {domain_bucket}")
         if bucket and domain_bucket and bucket != domain_bucket:
             msg = f"get_metadata_obj for domain: {obj_id} but bucket param was: {bucket}"
             log.error(msg)
@@ -334,7 +335,9 @@ async def get_metadata_obj(app, obj_id, bucket=None):
         if not bucket:
             bucket = domain_bucket
 
-    if not bucket:
+    if bucket:
+        log.debug(f"get_metadata_obj - using bucket: {bucket}")
+    else:
         log.warn("get_metadata_obj - bucket is None")
 
     # don't call validateInPartition since this is used to pull in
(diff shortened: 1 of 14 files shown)


hsds/util/domainUtil.py
```diff
@@ -51,22 +51,103 @@ def isIPAddress(s):
...
+        return True
+
+    bucket = _stripProtocol(bucket)
+
+    # Bucket names must contain at least 1 character
+    if len(bucket) < 1:
+        is_valid = False
+
+    # Bucket names can consist only of alphanumeric characters, underscores, dots, and hyphens
+    # other than
```
mattjala, line 92: This comment looks to be a leftover snippet

hsds/datanode_lib.py
```diff
@@ -327,14 +327,17 @@ async def get_metadata_obj(app, obj_id, bucket=None):
     log.info(f"get_metadata_obj: {obj_id} bucket: {bucket}")
     if isValidDomain(obj_id):
         domain_bucket = getBucketForDomain(obj_id)
+        log.debug(f"tbd - domain_bucket: {domain_bucket}")
```
mattjala, line 330: Dev comment should be removed

mattjala (dismissed):
Besides nitpicks, LGTM

hsds/datanode_lib.py
jreadey: done

hsds/util/idUtil.py
```diff
@@ -21,6 +21,43 @@
 from .. import hsds_logger as log
 
 
+S3_URI = "s3://"
+FILE_URI = "file://"
+AZURE_URI = "blob.core.windows.net/"  # preceded with "https://"
+
+
+def _getStorageProtocol(uri):
+    """ returns 's3://', 'file://', or 'https://...net/' prefix if pressent.
```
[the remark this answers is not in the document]
jreadey: fixed

PR #340, opened 2024-04-08
Update scripts to use docker compose v2

### LINKED ISSUE #339: Scripts use deprecated docker-compose v1
Docker Compose v1 is deprecated and has been [removed from the default github runners](https://github.com/actions/runner-images/issues/9557), leading to [failures with workflows that use HSDS scripts](https://github.com/HDFGroup/hdf5/actions/runs/8553582219/job/23437064298).

This should be as simple as replacing `docker-compose` with `docker compose` on the command line to force usage of the updated compose on runner machines.

mattjala:
Docker Compose v1 is deprecated - see #339.

This replaces CLI invocations of `docker-compose` with `docker compose`, which uses v2 instead.

To test the new scripts, this adds a CI workflow to build and test with Docker. Docker + Windows still has some issues, so tests with that combination are skipped for now. I also moved h5pyd tests from being a matrix option to the default, since that doubled the number of workflows for no good reason.

Results of workflow run can be seen [here](https://github.com/mattjala/hsds/actions/runs/8606446380).

Resolves #339.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -8,6 +8,9 @@ on:
   pull_request:
     branches: [master]
 
+env:
+  ROOT_DIR: ${{github.workspace}}/data
+
 jobs:
   build-and-test:
     strategy:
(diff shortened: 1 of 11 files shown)


jreadey (approved):
Looks good.
I followed the instructions on a clean Amazon Linux 2023 system and everything worked.

PR #342, opened 2024-04-09
Return internal server error on failed hyperslab write

mattjala:
- Remove `testPutVLenUTF8`, since the case it was testing (sequence of variable-length strings) isn't supported by HSDS.
- `_doHyperslabWrite` now causes the client to receive an internal server error if it fails, instead of silently failing and still returning 200.


jreadey (approved):
lgtm

mattjala:
Test should be restored and renamed to make it clear it's just checking that HSDS doesn't crash

jreadey (approved):
looks good!

PR #346, opened 2024-04-12
Fix inaccurate creation order on Windows

### LINKED ISSUE #344: Creation timestamps lack resolution on Windows
`time.time()` is used to get the current time when creating a link or other object. This works fine on POSIX systems, but on Windows only has a precision of about ~16ms. This means that two subsequently created objects can have the same creation time. Sorting links by creation order no longer consistently returns the link that was actually created first.

This issue shows up when testing `H5Lget_link_name_by_idx` through the REST VOL on Windows - see [here](https://github.com/HDFGroup/vol-rest/blob/1930a61102556d71c11e3237463e26871c576dad/test/test_rest_vol.c#L12796).

We could potentially use an external library for getting the UNIX time to have enough precision on all systems.

mattjala:
On Windows, `time.time()` has a low resolution (~16ms), so objects created in sequence sometimes have the same 'created' timestamp.

Creation time is now computed by adding the precise time elapsed since node startup (from `time.perf_counter`) to the recorded unix time at node startup, in the new helper `timeUtil.getNow()`.

Fixes #344

### DIFF
--- a/hsds/async_lib.py
+++ b/hsds/async_lib.py
@@ -10,7 +10,6 @@
 # request a copy from help@hdfgroup.org.                                     #
 ##############################################################################
 
-import time
 import hashlib
 import numpy as np
 from aiohttp.client_exceptions import ClientError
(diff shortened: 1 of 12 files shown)


jreadey (approved):
This looks good.
Potentially there is still an issue comparing timestamps set by different nodes (say the creation time of two different datasets).  If they are instantiated on different DNs then there's no guarantee the time values correlate with the actions of the client.  But since this only comes up in link creation comparisons where the links are on a single object it should be fine.

PR #351, opened 2024-04-19
remover docker compose version

jreadey:
Remove version key work in docker compose to avoid deprecation warning.  See: https://github.com/compose-spec/compose-spec/blob/master/spec.md#version-top-level-element.


mattjala (approved):

jreadey:
There was one test failure on windows (domain_test.py testCreateDomainNameIds).  
Looks like #344  is still a problem.

PR #352, opened 2024-04-22
remove dns style for domains

jreadey:
Removing support for dns-style domain names.
Previously a domain path like: "/home/test_user1/tall.h5" could be expressed as:
"tall.test_user1.home".
The DNS style derived from the original rest HDF white paper: https://www.hdfgroup.org/pubs/papers/RESTful_HDF5.pdf,
but this doesn't seem to have been used in practice or have any special utility over the more common style.  Also, there's some ambiguity with dots that are used as part of the filename rather than a folder separator.

### DIFF
--- a/hsds/util/domainUtil.py
+++ b/hsds/util/domainUtil.py
@@ -228,7 +228,6 @@ def validateDomain(id):
         raise ValueError("Slash at end not allowed")
     # the non-bucket part of the domain should start with '/'
     domain_path = getPathForDomain(id)
-    print("domain_path:", domain_path)
     if not domain_path or domain_path[0] != "/":
         raise ValueError("Domain path should start with '/'")
 
--- a/hsds/util/domainUtil.py
+++ b/hsds/util/domainUtil.py
@@ -226,6 +226,11 @@ def validateDomain(id):
         raise ValueError("Domain names should include a '/'")
     if id[-1] == "/":
         raise ValueError("Slash at end not allowed")
+    # the non-bucket part of the domain should start with '/'
+    domain_path = getPathForDomain(id)
+    print("domain_path:", domain_path)
+    if not domain_path or domain_path[0] != "/":
+        raise ValueError("Domain path should start with '/'")
 
 
 def isValidDomain(id):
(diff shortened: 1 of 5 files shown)


hsds/util/domainUtil.py
```diff
@@ -226,6 +226,11 @@ def validateDomain(id):
         raise ValueError("Domain names should include a '/'")
     if id[-1] == "/":
         raise ValueError("Slash at end not allowed")
+    # the non-bucket part of the domain should start with '/'
+    domain_path = getPathForDomain(id)
+    print("domain_path:", domain_path)
```
mattjala, line 231: Print should be removed

tests/integ/domain_test.py
```diff
@@ -46,20 +46,6 @@ def testBaseDomain(self):
         self.assertEqual(rsp.status_code, 200)
         self.assertEqual(rsp.headers["content-type"], "application/json; charset=utf-8")
 
-        # verify that passing domain as query string works as well
-        del headers["X-Hdf-domain"]
-        req += "?host=" + self.base_domain
-        rsp = self.session.get(req, headers=headers)
-        self.assertEqual(rsp.status_code, 200)
-        self.assertEqual(rsp.headers["content-type"], "application/json; charset=utf-8")
-
```
mattjala, line 55: Lines 49-54 should be able to stay, since this part is just testing that a regular (non-DNS) domain as a query string works.

tests/unit/domain_util_test.py
```diff
@@ -58,6 +57,7 @@ def testValidDomain(self):
 
         valid_domains = ("/gov/nasa/nex", "/home", s3_path, file_path, azure_path)
         for domain in valid_domains:
+            print(domain)
```
mattjala, line 60: Print should be removed

mattjala (dismissed):
LGTM

hsds/util/domainUtil.py
jreadey: fixed

mattjala (approved):

PR #354, opened 2024-04-22
Fix 409 Conflict in domain test on Windows

### LINKED ISSUE #353: 409 Conflict  during testCreateDomainNodeIds
Observed in the Windows, Python 3.12, no docker workflow:

```
======================================================================
FAIL: testCreateDomainNodeIds (__main__.DomainTest.testCreateDomainNodeIds)
----------------------------------------------------------------------
  File "D:\a\hsds\hsds\tests\integ\domain_test.py", line 531, in testCreateDomainNodeIds
AssertionError: 409 != 201
```

mattjala:
`testCreateDomainNodeIds` sometimes gets an HTTPConflict trying to create its domain. This is because the domain helpers generate a domain name based on the current timestamp and the given domain name. Both this test and `testCreateDomain` use the same given domain name of `newdomain.h6`, and because of Windows' low time resolution, sometimes both domains are created at the same timestamp and have a name conflict. 

This PR changes the name of the domain in `testCreateDomainNodeIds` to avoid the conflict regardless of timestamp.

Resolves #353

### DIFF
--- a/tests/integ/domain_test.py
+++ b/tests/integ/domain_test.py
@@ -521,7 +521,7 @@ class DomainTest(unittest.TestCase):
             self.assertEqual(root_id, rspJson["root"])
 
     def testCreateDomainNodeIds(self):
-        domain = self.base_domain + "/newdomain.h6"
+        domain = self.base_domain + "/newdomain.h7"
         print("testCreateDomainNodeIds", domain)
         headers = helper.getRequestHeaders(domain=domain)
         req = helper.getEndpoint() + "/"


derobins (approved):

jreadey:
Curious - why did this only crop  up on Windows?

mattjala:
> Curious - why did this only crop up on Windows?

The root issue is the same as with most of the other Windows issues recently: `time.time()` only has a resolution of about 16ms, so it's not uncommon for two objects to be created within the same <16 ms period of time and be given the exact same timestamp.

PR #355, opened 2024-04-22
Fix scan timeout on Windows

mattjala:
After #349, Windows still sometimes has a scan timeout due to `getNow` inaccuracy between nodes. This reverts the scan timestamps to use `time.time()`. Because the [scan check](https://github.com/HDFGroup/hsds/blob/master/hsds/domain_sn.py#L916) now accepts an equal timestamp as satisfying the scan request, the inaccuracy of `time.time()` shouldn't be a problem in this case.


jreadey (approved):
looks good

PR #356, opened 2024-04-24
Add link description to schema doc

mattjala:
Links seem to be stored as sub-objects in the same manner as attributes, so it makes sense to describe them here. 

Also fixes some typos.


jreadey (approved):
lgtm

PR #357, opened 2024-04-26
Enable codespace for hsds

jreadey:
This PR enables the hsds repo to be open in a github codespace.
The test suite (python testall.py) should work when run within the codespace.


mattjala:
When running the tests from a codespace, the point selection test keeps failing due to an import issue:
```
pointsel_test
  File "/workspace/tests/integ/pointsel_test.py", line 18, in <module>
    from hsds.util.arrayUtil import arrayToBytes
ModuleNotFoundError: No module named 'hsds'
Failed
```

data/hsdstest/home/.domain.json
```diff
@@ -0,0 +1 @@
+{"owner": "admin", "acls": {"admin": {"create": true, "read": true, "update": true, "delete": true, "readACL": true, "updateACL": true}, "default": {"create": false, "read": true, "update": false, "delete": false, "readACL": false, "updateACL": false}}, "created": 1708897646.0599918, "lastModified": 1708897646.0599918}
```
mattjala, line 1: Looks like the `data` folder was included by mistake

mattjala (approved):
Besides including the `data` folder, LGTM

jreadey:
I hadn't setup the codespace environment with the hsds package, so that caused the failure.
Seems like this is the only integ test using the hsds package, and it's not really needed anyway.  I just replaced the arrayToBytes() call with arr.tobytes() and everything passed.

PR #360, opened 2024-05-20
Fix some incorrect pattern parsing

mattjala:
Pattern with asterisks at the end are incorrectly considered matches, even if the preceding pattern does not match


jreadey (approved):
lgtm

PR #363, opened 2024-05-22
Use old requests version

mattjala:
The requests 2.32.0 update from a couple days ago breaks requests-unixsocket for connections to a local socket. There is [a PR](https://github.com/msabramo/requests-unixsocket/pull/72) to requests-unixsocket, but until that gets in we should constrain the requests version


jreadey (approved):
lgtm

PR #373, opened 2024-06-11
fixes for quick start usage

jreadey:
Updates so quickstart instructions work


mattjala (approved):
LGTM. The test failure is due to a 409 conflict trying to create a dataset during a test on Windows. I suspect it's another result of two domains having the same name due to low timestamp resolution - either way, nothing related to this set of changes.

mattjala:
It looks like these changes are also included in #372. They should either be dropped from that PR, or this should be closed

mattjala (approved):

PR #371, opened 2024-06-10
Listkeysv2

jreadey:
Use v2 interface for list keys (compatible with S3 express buckets)

PR #374, opened 2024-06-12
Fix socket connections

mattjala:
After #371, socket operations started failing. This seems to be due to an incompatibility between `requests>=2.29.0` and `urllib3>=2.0.0`. The only change that PR made to dependencies was to bump `aiobotocore`, which doesn't require `urllib3>=2.0.0`, so I'm not sure why the package versions downloaded changed as a result.

To resolve this
- fix the version of `urllib3<=2.0.0` to avoid the incompatibility with newer requests
-  Also fix `requests <=2.31.0` to fix the incompatibility with `requests-unixsocket`. `The requests version was already fixed at this value in #363, but it looks like a dependabot PR was merged after that.

This PR also includes a new workflow to test the socket specifically to prevent this from happening in the future.


jreadey (approved):
looks good!

PR #377, opened 2024-06-18
Prevent use of new major numpy version

mattjala:
Numpy just had its [first major release in years](https://github.com/numpy/numpy/releases/tag/v2.0.0), with breaking changes to its API. We should eventually  add compatibility, but for now just restrict the numpy dependency to a pre-2.0.0 version.


jreadey (approved):
lgtm

PR #389, opened 2024-08-26
Set platform in docker-compose so it works on other platforms than amd64

rho-novatron:
With this change, runall.sh works on e.g. macos with Apple Silicon.

### DIFF
--- a/admin/docker/docker-compose-internal-lb.yml
+++ b/admin/docker/docker-compose-internal-lb.yml
@@ -1,5 +1,6 @@
 services:
   head:
+    platform: "linux/amd64"
     image: hdfgroup/hsds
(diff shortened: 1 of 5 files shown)


rho-novatron:
The test fail is this:

’’’
  File "D:\a\hsds\hsds\tests\integ\value_test.py", line 3809, in testShapeReinitialization3D
AssertionError: 409 != 201
’’’

I can’t really see how this change could provoke that test fail…

jreadey (approved):
Looks good!  Thanks for the PR .
The test fail is a known issue on the windows platform.  Since the clock resolution on Windows is less compared with Linux/MacOS, tests the depend on listing attributes or links by creation order sometime fail.

PR #404, opened 2024-12-11
fix logging deprecation warning

jreadey:
Replace logging.warn with logging.warning


jreadey:
update deprecated logging.warn call

PR #408, opened 2024-12-12
Fix CI failures due to domain name conflits

mattjala:
- Prevent domain name conflict during tests

Tests use the timestamp to generate unique domain names and avoid conflicts. On windows, the timestamp resolution is low enough that these sometimes have the same timestamp and still conflict with a 409 error.

This replaces the timestamp with an 8-character random string, which should have less than a one in a million chance of a name conflict.

We should probably remove the duplicated test helpers at some point, but it seems like setting up a shared import between the tests that would use it is non-trivial

### DIFF
--- a/tests/integ/helper.py
+++ b/tests/integ/helper.py
@@ -13,14 +13,9 @@ import requests
 import requests_unixsocket
 import json
 import os.path as op
-from datetime import datetime
-import time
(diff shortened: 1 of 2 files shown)


jreadey (approved):
lgtm

PR #427, opened 2025-06-11
Replace deprecated numpy fromstring()

mattjala:
Numpy 2.3.0 deprecated the default (binary) mode of `np.fromstring()`. This replaces use of `np.fromstring()` with `np.frombuffer()`, which should have no change in behavior.


jreadey (approved):
looks good!

PR #430, opened 2025-06-30
Replace unix socket package

mattjala:
- Replace deprecated unixsocket dependency with a fork that is still maintained

The original package is incompatible with requests 2.32.0+.

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -213,11 +213,6 @@ jobs:
         run: |
           pip install -e .
 
-      # Requests 2.32.0 breaks requests-unixsocket, used by HSDS for socket connections
-      # - name: Fix requests version
-      #  run: |
-      #    pip install requests==2.31.0
-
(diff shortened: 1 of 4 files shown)


mattjala:
Botocore and the replacement for unixsocket have an unavoidable dependency conflict for Python 3.9. Options for resolving this are to drop support for Python 3.9, freeze requests/urllib/requests-unixsocket at their current version, or replace requests-unixsocket with something else

jreadey:
I'd vote to drop support for Unix Sockets.  Originally I was thinking that this might be a performance enhancement for clients and servers on the same host, but it's marginal if anything.  Also, I think there can be problems when two clients are using the same socket connection.

If we go this way, the  ruanll.sh script should change the --no-docker option to use TCP rather than sockets.

jreadey:
Actually we use unix sockets in hsds_app.py, and I prefer not to update the code to use tcp sockets.  
Let's go with changing the unix-socket to use the gitlab project.

